PR #28, opened 2017-07-11
Features/etrago function

wolfbunke:
Please check the new Feature of eTraGo as function.

call: network = etrago(args)


wolfbunke:
thanks

PR #38, opened 2017-09-05
SettingWithCopyWarning in appl.py

s3pp:
Executing appl.py I got several SettingWithCopyWarnings. I tried to fix it with the following code.


s3pp:
Nice, thanks for reviewing!

Issue #23, opened 2017-07-03
Calculate grid loss costs with PF


lukasol:
See #10 for the implementation of the PF. 
Especially with regards to the eGo tool, we need full costs for grid losses.

- [x] Calculate the grid losses for all lines using the results of the PF
- [x] assume average costs to cover these losses

lukasol:
Consider fundamentals of electricity...not just the simple diffs because with pf Q parts lead to increased flows.

lukasol:
We'll mainly focus on the copper losses `P = I^2 x R` for all lines, calculated with the PF results.

Additionally, voltage-based losses for the transformers could be assumed by using the standard losses given by the producers for a specific set of pre-defined transformers. These are not dependent of the current and can thus easily be defined.

Other losses will be disregarded.

ulfmueller:
as a base you can use the [pf_after_lopf branch](https://github.com/openego/eTraGo/tree/features/pf_after_lopf) for your loss calculation. It works very fine for the SH status quo scenario and for the timesteps chosen. But it seems to work fine with basically all snapshots. The problem is the bug in the x of the transformer which changes a lot. In the new dp version the x will be fine. In order to fix this bug temporally there is a line which downgrades the x of the tranformers at least in the order of magnitude where it should be (still being conservative). 
Anyhow, bottom line it generates pf results which you can use for loss calculation!

lukasol:
I added the function `calc_line_losses` in https://github.com/openego/eTraGo/commit/eb4830805027b5ea62ec28a5a52096896a066ac4

@ulfmueller please check!

ulfmueller:
Looks good. Havent tried it out yet. 
Which could be added is the calculation of the transformer copper losses. The copper losses of transformers can be calculated by
loss_tranformer = (apparent power S/S_nom)² * nominal copper losses of the transformer. There as well we have to make assumptions for our nominal copper losses...

lukasol:
> loss_tranformer = (apparent power S/S_nom)² * nominal copper losses of the transformer. There as well we have to make assumptions for our nominal copper losses...

That would mean that we do consider power-dependent losses of transformers. We could also just use the no-load losses according to the producers. Would you know of a nice source for this?

lukasol:
The general logic has been implemented in #41 

Open issues:

- [ ] calculate copper losses for transformers
- [ ] find a nice way to include the results of the loss calculations to the network container

lukasol:
[...]

Issue #10, opened 2017-05-24
pf() after lopf()


ulfmueller:
- [x] develop a functionality to couple lopf and pf in order to simulate Q and line losses after having done the linear optimization.
- [x]  implement the cosphi assumptions based on the lopf results
- [x]  implement and consider slack bus
- [ ] test the PF results: are there feasible results for DE and does the consideration of Q and losses lead to line loading > 100 %?
- [x] (calculation of additional costs should be covered by #23 ) 
- [x] work on a feature branch
- [x]  define a function

ulfmueller:
added a branch with a first coupling lopf pf example: https://github.com/openego/eTraGo/blob/features/pf_after_lopf/etrago/appl.py#L106-L138

lukasol:
Just for your info: I just did a LOPF - then - PF run for the hours 1 and 2 for Germany NEP 2035 and it solved.

lukasol:
...but the error of the PF is not "0.000", but "nan"...

ulfmueller:
The functionality is already merged into dev branch [see here](https://github.com/openego/eTraGo/blob/dev/etrago/extras/utilities.py#L195) and it can be chosen in the appl.py. The feature branch is still there because there still some features (see check list above) pending.

lukasol:
What exactly do you mean by additional costs? The costs induced through transmission losses? That would be part of #23

ulfmueller:
Your are right. That should be it and part of #23 . I changed it but left it there as a link, to have it in mind in case I forgot an effect. But I think we do not have to consider other costs due to the non-linear pf.

ulfmueller:
I forgot s.th. so far. We have to consider the position and usage of the slack in the pf!
For the position we defined it in redmine  #935. It is still pending to implement this into this function.
Also an evaluation of the usage of this slack should be undertaken. We have to tackle possible inbalances in an adequate way.

ulfmueller:
The implementation of the cosphi assumptions is trivial. all big pp are PV generator. all small pp have a cosphi=1. (assumptions see redmine #935). Therefore we do not need any q_set for generators. 
We just have to be sure that the q_set from the loads are imported. That should be the fact looking at the [json](https://github.com/openego/ego.powerflow/blob/dev/egopowerflow/tools/config.json#L13)

ulfmueller:
@mariusves will implement the slack bus definition.

mariusves:
[...]

PR #41, opened 2017-09-07
Features/pf after lopf

lukasol:
Include the general logic of loss calculation for lines and transformers according to #23 
There's still something to enhance, so this is a preliminary version.


lukasol:
I leave the branch open for #23 and #10

PR #45, opened 2017-09-08
Release 0.3

lukasol:


lukasol:
Right, forgot that!

ulfmueller:
mhh. now it links to the new ego.pf release which is still not submitted to pypi. which would mean installation will not work like that. should we wait for that? Or do we want to change the setup.py?

ulfmueller:
Forgot what I said. I missed that it was release already.

PR #58, opened 2017-09-26
Feature/dissolve ego powerflow in etrago

mariusves:
With this, all functionalities of ego.powerflow would be integrated into eTraGo. 
Since this merge of the two packages didn't require any fundamental changes, there should be no further issues with this. At least when I tested it I couldn't find any bugs or references to the old ego.pf in other modules of eTraGo apart from the appl.py.


lukasol:
Sounds very reasonable!

ulfmueller:
one more thing. When this is approved please consider what has recently been developed in ego.powerflow such as the plotting adjustment by @MarlonSchlemminger done in this [branch](https://github.com/openego/ego.powerflow/tree/features/line-loading-plot/egopowerflow/tools). As far as I know that is the only thing going on in ego.powerflow after you did this PR...

ulfmueller:
@mariusves could you change the repos structure and mind the recent changes to ego.powerflow?
And then we merge this! It looks like that we can do the merge without waiting for the release since the dev keeps on working without changing the requirements (which makes somehow sense).

mariusves:
Sure, I can do the restructuring as suggested and also implement Marlons work. @MarlonSchlemminger how far is the development of that branch? Can I take the plot.py to be implemented into eTraGo?

@ulfmueller Another question about the [snapshot_clustering.py in the folder "plots"](https://github.com/openego/eTraGo/blob/dev/etrago/plots/snapshot_clustering.py) and the [snapshot_clustering_app.py](https://github.com/openego/eTraGo/blob/dev/etrago/snaphot_clustering_app.py) in the main packages folder: Are these still being used or are they obsolete by now (since there is also a [snapshot.py](https://github.com/openego/eTraGo/blob/dev/etrago/cluster/snapshot.py) in the "cluster" folder.

ulfmueller:
those snapshot files are still relevant. Well, they are used from a feature branch - so the day we merge it back to dev it becomes relevant. Right now in the dev they do not have any usage...
But I think you can just leave them where they are or will be.

MarlonSchlemminger:
@mariusves You can take plot.py in that branch.

lukasol:
@mariusves could you give me a short hint once the restructuring mentioned above is done? I would give you the review then asap.

lukasoldi:
Don't we need a hint to pypsa also in setup.py (and not only in the requirements.txt)?

mariusves:
Package is now restructured as suggested. A hint to PyPSA in the setup.py should not be necessary, since when installing using the requirements.txt is essential anyways (eTraGo will not work correctly when not using the eGo fork of PyPSA). Unfortunately, the `dependency_links` in setup.py doesn't work (it's a PIP issue, already discussed on the pip git).

I tested the current state of this packages and so far I couldn't encounter any bugs. Can someone else please check, too?

mariusves:
PR and merge is now complete.

Issue #61, opened 2017-10-06
Check oedb session compatibility to ego.io connection


mariusves:
Taken from @ulfmueller (issue in ego.powerflow):

Check out how the [session maker in ego.eTraGo](https://github.com/openego/eTraGo/blob/dev/etrago/tools/utilities.py#L22-L48) works with the adjusted [session maker in ego.io](https://github.com/openego/ego.io/pull/24/commits/4c2caea68c7da5e633d1190b7b456d3eb8013e34). 

- [x] Check if we can get rid of the function here in ego.eTraGo and do that only in ego.io centrally.
- [x] may be create a functionality that it is also possible to insert the session parameters manually

mariusves:
With commit 52b9cd1873b5264d883ef7d74b71d3dc2f293e15 I implemented the use of the ego.io sessionmaker, which also allows to use the function, that creats the ego.io-config file.
The function `oedb_session` now returns two objects: `session` and `conn`. While the former is the already in use session-object, the latter will be later used to extract the username when exporting to the oedb, so results can be easily linked to the person who uploaded them.

mariusves:
I did some testing, and found the ego.io sessionmaker quite wonky. 
When there is already 1 server stored in the config file, no further server can be added. Furthermore, passwords stored in the keyring can't be retrieved which means everytime you run the tool, you have to re-enter the password. I'm not sure if this is a specific windows related problem, since I don't have access to a linux computer right now. Can someone please test that?

PR #92, opened 2017-11-15
change of db connection

wolfbunke:
The automatic db connection with oemof.db was difficult to setup. There for I add this function of @s3pp for an easier DB connection.
 
- remain of DB configuration settings in .oemof/config.ini
- remove oemof.db
- add dbconnect and readcfg function

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -188,7 +188,7 @@ def etrago(args):
 
     """
 
-    session = oedb_session(args['db'])
+    session, conn = oedb_session(args['db'])
 
     # additional arguments cfgpath, version, prefix
     if args['gridversion'] == None:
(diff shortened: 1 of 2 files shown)


mariusves:
This update seems to be a kind of duplicate of issue #61, but a better one. I wasn't quite happy with the way the config handling worked in the ego.io/oemof.db functions. 
Your way seems to be working better. I added the function `writecfg` which kicks in if there is no config.ini in the folder `.open_eGo`, or if there is a certain section missing. 

Since I could only test this on windows, could you please test if this works on Linux as well? Sometimes the `os.makedir` functions are behaving differently on windows and linux.

mariusves:
After discussing with @ulfmueller we agreed to use the new functions of this branch to be used in the ego.io sessionmaker, as ego.io already acts as a central coupling tool between the python environment and the database.
I reopened issue 16 of ego.io [here](https://github.com/openego/ego.io/issues/16) where the new sessionmaker was introduced and added some points for discussion there. This PR should therefore not be merged.

wolfbunke:
ok I will close it now.

Issue #77, opened 2017-10-18
Too few clusters in germany


MGlauer:
Using k-Means clustering with *NEP 2035* yields 9 Clusters. The reason seems to be that all but 9 weightings are set to 0 - thus only 9 Buses are distributed in 100 Clusters.

Is this linked to the [missing `conv types` in non-SH scenarios](https://github.com/openego/eTraGo/blob/7754b410a8d5401e211a67a6c44c5120ef549674/etrago/cluster/networkclustering.py#L325) or does it point to a deeper bug?

ulfmueller:
That's right. I got the same problem. Good question. I will try out if it has to do with the inflexible weakness you pointed out.
@wolfbunke Can you come up with a more generic solution which also works for other non-SH scenarios such as our 'normal' scenarios?

ulfmueller:
I checked it and the result is as I expected. This problem does not have to do with the conv types. So it seems to be another bug, hopefully not a deeper one. Can anyone look for the reasons? @wolfbunke @BartelsJ ?

wolfbunke:
@BartelsJ  could you take a look to this problem?

wolfbunke:
Thank you @ulfmueller !

BartelsJ:
Sure, I'm on it

ulfmueller:
After @BartelsJ told me that the same problem occurs when applying clustering to the SH scenario I checked my simulations on SH a couple month ago. They were still fine concerning the k. After thinking about what had changed in the meantime I switched back to the old simple weighting using these [lines](https://github.com/openego/eTraGo/blob/dev/etrago/cluster/networkclustering.py#L399-L400) instead of the more sophisticated method being called [here](https://github.com/openego/eTraGo/blob/dev/etrago/cluster/networkclustering.py#L401-L402).
Turns out, then it worked fine again for all scenarios. So the problem lies within the more sophisticated weighting method implemented by @wolfbunke

wolfbunke:
@ulfmueller thanks for checking the methods. I will take a look into it and try to find the buggy part.

wolfbunke:
After checking the kmean in  0cbaa1b by using the `SH Status Quo` data from v02.11 with ` 'start_snapshot': 1`, ` 'end_snapshot' : 24` as well as from time step 2000 to 2020. 
I found the problem by many equal (zero) values in generation and different numbers of load and generator buses. This data set has original 91 buses of which 14 has a p_nom value over zero and 21 load buses with values. 

[...]

PR #93, opened 2017-11-15
change weighting factor from 100 to 100000

wolfbunke:
@ulfmueller and me found the problem by the factor of the weighting function
A more generic solution is needed
see #77


ulfmueller:
I would leave the branch open, then you can develop the more generic solution there, ok?

wolfbunke:
I implemented a more generic solution here   ef4fb2f   with
-  args description change of kmean
- w_method as filter for the weighting
- a generic solution to find a good factor for the weighting function

PR #94, opened 2017-11-15
Upgrade to pypsa 11

ClaraBuettner:
In eTraGo/io I changed the version number of pypsa, so components can be renamed and in networkclustering.py I changed network.now, that has been revomed in pypsa 0.9, to network.snasphots.
I tested it local and it worked.


ClaraBuettner:
Thats why I changed it in the pypsa-11-branch to network.snapshots, which worked well.

PR #99, opened 2017-11-22
merge brunch Fix#77 into features/Kmean#77

wolfbunke:
Merge a more generic function to calculate the weighting factor

ToDo: 

- [x]  @wolfbunke make testing of run time of features/Kmean#77 and function of Fix#77  before merging


wolfbunke:
The run time of both method is more or less equal. I was testing each setting three times and the difference  was between 0 and 3 sec. (49 to 52 sec)

wolfbunke:
solved merge conflicts

PR #129, opened 2018-02-12
Add shaply to install_requires setup.py

wolfbunke:
The python package shapely is missing in setup.py  install_requires.
After installing and using eTraGo in a clean environment for a single usage an import error of the missing package shapely appears. This should fix the error.


ulfmueller:
@wolfbunke af1525bc46407bec23e3870dd251c39a7ef815a2 added shapely on the dev. But without a specific mentioning of version numbers. Is that sufficient and we close this PR?

wolfbunke:
I guess we can close it

PR #132, opened 2018-02-21
Add shapely to dependency list

gplssm:
Shapely is a missing dependency. Do you want to include it in the list of depedencies?


ulfmueller:
Thx. Sounds good!

PR #151, opened 2018-04-23
add remove stubs function to network clustering and documentation

ulfmueller:
The k-means network clustering was not doing the first step of the methodology such as in [Hoersch et al.](https://arxiv.org/pdf/1705.07617.pdf) in order to remove dead-ends. I implemented the function using basically the pypsa function. It is optional and can be set to True if wanted. 
Please try it out and see if it works fine for you. One review is enough.

Moreover I added some documentation and did some line breaking which does need further review, I think.


ulfmueller:
@BartelsJ could you please give your review so I can merge!?

Issue #31, opened 2017-07-17
Use extendable lines to replace load shedding


lukasol:
Depending on the results of load shedding analyses in #19 allow the respective lines (those that connect nodes with increased load shedding) to be extendable. Then test this functionality and its potential to replace load shedding in a new run.

lukasoldi:
@kimvk could please state your current progress briefly here? Thanks!

kimvk:
When I allow all lines to be extendable, load shedding disappears.
If I extend only the lines connected to the nodes with load shedding, load shedding is lowered, but most load shedding spots will remain. 

Next step will be to extend the lines which are connected to the lines extended before until load shedding disappears. This will slower the computation time, because after each extension of the lines, the LOPF will be calculated again to see if load shedding still exist.

ulfmueller:
maybe we do not have to be that complicated. How does the solution look when you allowed all lines to be extended? Can we may be use this solution as the desired 'debugged' data solution?
Is there a huge grid expansion or only a slight one concerning only the effect of load shedding? Does the dispatch change dramatically or only slightly?
Maybe we can fix the generator dispatch of the load shedding lopf with the help of p_min_pu to limit a substantial change of dispatch and only tackle the load shedding.

Or even better you just set the capital cost of the lines to a very, very expensive value - that it wont change the dispatch substantially and only fix the problems caused by load shedding (load shedding turned off of course) because it has to supply the load. I think that is a very good idea. @kimvk Can you try this out? It should be a fast, easy and nice solution.

Sorry for doing a simultaneous thinking-writing and letting you read my thoughts... I think you can read right away the last paragraph.

kimvk:
I pushed the new code. I set all lines and transformers extendable and the capital cost to 1 million. The results of s_nom_opt for the lines and the transformers are saved in two csv fils, which can replace the old s_nom in a new calculation. If we calculate without loadshedding and with the data from the csv files the computation time is essentially faster.

MarlonSchlemminger:
I've pushed a new version of code. Two csv-Files (one for transformers, one for lines) including all components with changed s_noms are produced and can then be used to overwrite s_noms in the next run. In a new environment on the server with most recent versions of etrago and egoio, it looks like there are no unbounded timesteps anymore (at least some timesteps which were unbounded before are now feasible). This means we can probably run this as soon as the dp run is finished.

MarlonSchlemminger:
[...]

PR #154, opened 2018-05-08
improve topology in Munich

jankaeh:


ulfmueller:
I will check your approach once our new data release 0.4.2 is finished and see how we can combine it with #31 This will happen the latest by the end of July.

lukasoldi:
I tried applying your approach, but got this error. Any clues on that?

> ~/eTraGo/etrago/appl.py in etrago(args)
>     313 
> --> 314     add_missing_components(network)
>     315 
>     316     # grouping of parallel lines
> 
> ~/eTraGo/etrago/tools/utilities.py in add_missing_components(network)
>     360      ) = (from_shape(MultiLineString
>     361                      ([LineString([wkb.loads(network.buses.geom['23648'],
> --> 362                                              hex=True),
>     363                                    wkb.loads(network.buses.geom['16573'],
>     364                                              hex=True)])]), 4326))
> 
> ~/lukas_ego/lib/python3.5/site-packages/shapely/wkb.py in loads(data, hex)
>      12     reader = geos.WKBReader(geos.lgeos)
>      13     if hex:
> ---> 14         return reader.read_hex(data)
>      15     else:
>      16         return reader.read(data)
> 
> ~/lukas_ego/lib/python3.5/site-packages/shapely/geos.py in read_hex(self, data)
>     404             data = data.encode('ascii')
>     405         geom = self._lgeos.GEOSWKBReader_readHEX(
> --> 406             self._reader, c_char_p(data), c_size_t(len(data)))
>     407         if not geom:
>     408             raise WKBReadingError(
> 
> TypeError: bytes or integer address expected instead of Function instance`
>

jankaeh:
Works on my system, but historically grown I don't use an entire installation of etrago, so I cannot really compare. Maybe you can just adopt the line from the function data_manipulation_sh, which should do the same without using wkb.loads, which did not work on my system though:
instead of:
from_shape(MultiLineString([LineString([wkb.loads(network.buses.geom['23648'],hex=True),wkb.loads(network.buses.geom['16573'],hex=True)])]), 4326))
it would be:
from_shape(MultiLineString([LineString([to_shape(network.buses.geom['23648']),to_shape(network.buses.geom['16573'])])]),4326))

same for topo and again at line 408-419

lukasoldi:
I have changed your function to comply with the full etrago set, see https://github.com/openego/eTraGo/commit/210a5d5b2a7d8802d0212d0a22d06af5e402a9fc

@jankaeh how did your result improve/change with the changes in place? In my case there is only a very slight change in results.

ulfmueller:
The adapted function is also part of the branch feature/test_powerflow which will be merged into dev soon. So I close this PR for now. 
Btw: Our experience was similar to @lukasoldi. Also when calculating the PF, our problems were coming from bugs and the abstract modelling of the neighboring countries instead these topology changes had a rather small impact. The PF is converging also without these changes. Anyhow the function will be included and can be used if one desires. Thx again @jankaeh for sharing your work!

PR #155, opened 2018-05-08
Features/snapshot clustering tsam single

ulfmueller:
this updates the snapshot clustering functionality being based on tsam.


ulfmueller:
@kimvk I just noticed that all the plotting functions are added here. I would delete the analyses folder...

Issue #157, opened 2018-05-09
bus weightings for kmeans


ulfmueller:
It makes sense to introduce a bus weighting which is calculated from the Status Quo scenario and use it also for the future scenarios. Therefore it would be good to implement a functionality to calculate and save a bus weighting and as well take a given weighting.

PR #160, opened 2018-05-17
Features/bus weighting

ulfmueller:
This solves #157


ulfmueller:
I already pulled from dev so it is mergable. After you have checked the functionality (maybe recheck remove_stubs also again) you can merge it yourself and delete the feature branch.

PR #173, opened 2018-05-29
Feature/change header

wolfbunke:
I changed the license text to a comment. Which may be better than removing it -  as discussed earlier.


BartelsJ:
Looks good! One remark: Shouldn't we also change this for the files networkclustering.py and snapshot.py in the etrago.cluster repository?

wolfbunke:
The pull request is now up to date with the changes of dev. Please have a look before I will merge it. @BartelsJ  @ulfmueller

wolfbunke:
Thanks for approving, I will merge it now.

PR #180, opened 2018-06-13
Features/plotting

ulfmueller:
The plots `storage_distribution` and `storage_expansion` got a legend now. Furthermore the attribute `scaling` can be used to set the sizing of the bubbles.

Furthermore a new plot was developed to plot the nodal generation over the entire period of time calculated as a pie per node where it can be specified which generation carrier one want to plot. This plot is still missing a legend due to this [issue](https://github.com/PyPSA/PyPSA/issues/32#issuecomment-383000409).

The `plot_line_loading` can now also plot the average loading over a period of time by specifying the range of snapshots you want to look at. To reproduce the old functionality of just plotting the results for one particular snapshot one has to state `timesteps=range(x,x+1)` instead of `timesteps=x`

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -71,7 +71,7 @@ args = {# Setup and Configuration:
         'reproduce_noise': False,# state if you want to use a predefined set of random noise for the given scenario. if so, provide path, e.g. 'noise_values.csv'
         'minimize_loading':False,
         # Clustering:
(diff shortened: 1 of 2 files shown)


lukasoldi:
Thanks, looks good! This raises the general question of the documentation of plotting functions. Should we do some more documentation within the respective functions?

ulfmueller:
Yes we should! 25th to 27th we will have 'intensive documentation day' together with @BartelsJ here in Flensburg. Hopefully we can tackle a lot then.

PR #194, opened 2018-07-06
Substitute noise_values.csv with seed number in generator_noise argum…

BartelsJ:
…ent.

Instead of stored and exchanged noise.values.csv files to reproduce and share the same random noise for generators, the generator_noise can be set to an arbitrary integer (seed number) which reproduces the same noise for the same integer.


ulfmueller:
I would say one review is enough for merging. Anyone disagrees? If not, @BartelsJ please merge...

PR #196, opened 2018-07-11
Bugfix/k means weighting

MarlonSchlemminger:


MarlonSchlemminger:
@ClaraBuettner I just realized that the commits I push from the student account appear under your name instead of mine, just in case you wonder where some of your commits come from. It is weird though, I always have to enter my login details when I push, I am not sure why it is under your name..

PR #211, opened 2018-08-07
Feature/rtd theoretical background

BartelsJ:
Update of eTraGos RTD theoretical background.


BartelsJ:
Thanks for the great remarks. I updated the description.

PR #225, opened 2018-08-29
Feature/nonlinear plots

ClaraBuettner:
Change plot_line_loading to consider links when solution is nonlinear


ClaraBuettner:
I adjusted the method to avoid covering of links in plots and it works now.

PR #234, opened 2018-09-04
Use solver_names, solver_options and extra_functionality in 2nd lopf

ClaraBuettner:


ClaraBuettner:
In addition, I deleted the cluster/analysis folder which was just needed when I tested some functionalities.

PR #241, opened 2018-09-11
Consider voltage level in branch capactiy factor function.

ClaraBuettner:


ClaraBuettner:
Yes, this is not used. I just copied the code from the set_trafo_cost_function. I'll delete these lines. 
The lower voltage is always bus0.

PR #256, opened 2018-10-23
Feature/storage plots

ClaraBuettner:


ClaraBuettner:
I basicly just copied the code from the paper branch. I'll test your args soon and try again.

ClaraBuettner:
I can't test with the same args you use because my RAM is too small for 500 buses. I tested with csv files and git other results. For the storage_soc I get scaled data, and I think the missing x-axis description is fine, because the values are sorted by value and not by the timeindex. 
For storage_stat and storage_p plot, my x-axis is split up in months, but I tested with a calcultaion of a year, maybe that's not working for small calculation. I'll try to run a calculation with less snapshots.

ClaraBuettner:
With less snapshots I get problems too. But I think that's not changeable; in storage_stat and storage_p plot time-series are resampled with high hours, so there is e.g. just one entry for 168h, which is not working when calculation only a few snapshots.

PR #266, opened 2019-02-04
Feature/variable constraints

ClaraBuettner:
This branch allows considering multiple extra_functionalities. In addition, limits of extra_functionalities are set in args of appl.py. 

The extra_functionality-function only works with parameters (network, snapshots) that's why I needed to create an object to allow setting parameters in appl.py.


etrago/appl.py
```diff
@@ -392,10 +405,10 @@ def etrago(args):
     network.storage_units.cyclic_state_of_charge = True
 
     # set extra_functionality
-    if args['extra_functionality'] is not None:
+    """if args['extra_functionality'] is not None:
```
ulfmueller: It does not seem very nice to keep these lines as a commentary , does it?

ClaraBuettner: Yes, you're right. I'll delete these lines.

ulfmueller:
I have another issue. Now the extra_functionalities are still defined as functions in the utilities.py. Wouldn't it be nicer if would only have them defined at one spot, hence at the new location constraints.py? Or what is the reason for still having them in the utilities?

ClaraBuettner:
The last commits allow to combine extra_functionalities with snapshot clustering.

PR #267, opened 2019-02-04
Include model_formulation parameter in args

ClaraBuettner:


ClaraBuettner:
I just merged the branches, see https://github.com/openego/eTraGo/pull/270

PR #268, opened 2019-02-14
Feature/pypsa13

ClaraBuettner:
I tested all etrago-functions (only oedb-export is missing currently, because it is very slow with vpn-connection) and adjusted some lines. 
The derivation in optimization results compared to the same etrago-parameters in dev-branch and with pypsa 0.11.0 are very small and in the same dimension of derivations in the result of the same setting. 
This branch needs [this](https://github.com/openego/PyPSA/tree/feature/pypsa_13_fork) pypsa-fork with version 0.13.2 and newer version of some packages (e.g. pandas > 0.23.0).

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -106,16 +106,16 @@ if 'READTHEDOCS' not in os.environ:
 
 args = {
     # Setup and Configuration:
-    'db': 'oedb',  # database session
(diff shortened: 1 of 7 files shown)


ClaraBuettner:
Since we will update to a newer version of pypsa I think we can close this PR.

PR #270, opened 2019-04-09
Feature/lopf iterations

ClaraBuettner:


ClaraBuettner:
I just pushed the changes we talked about: 
- line extension of the previous lopf is not set for the next iteration anymore
- maximum extension of each line/trafo ascends with iterations

ulfmueller:
looks fine. Just one aspect I would change. The max_ext I would keep line (component i.e. also transformer) specific, so if you just want to limit a few lines e.g. the cross-border lines only they would be effected instead of taking a mean over all lines...

ClaraBuettner:
In [this branch](https://github.com/openego/eTraGo/tree/feature/lopf_iterations_threshold) I added the possibility to choose if using a fixed number of iteration or a threshold.

ulfmueller:
Nice, can you make a PR from that branch to merge it into this one and add some documentation to the new functionality in the modified method?

ClaraBuettner:
This branch now includes iteration with threshold and model_formulation parameter in args

PR #271, opened 2019-05-27
Feature/update decommissioning

ClaraBuettner:
This branch allows reducing the capacity of all components using decommissioning scenarios.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -117,8 +117,8 @@ args = {
                        'logFile': 'solver.log', 'threads':4, 'method':2, 'crossover':0},  # {} for default options
     'scn_name': 'NEP 2035',  # a scenario: Status Quo, NEP 2035, eGo 100
     # Scenario variations:
-    'scn_extension':['nep2035_b2'],  # None or array of extension scenarios
-    'scn_decommissioning': ['test'],  # None or array of decommissioning scenarios
+    'scn_extension':['nep2035_b2', 'wind_offshore'],  # None or array of extension scenarios
+    'scn_decommissioning': ['bugfix_nep', 'nep2035_b2'],  # None or array of decommissioning scenarios
     # Export options:
     'lpfile': False,  # save pyomo's lp file: False or /path/tofolder
     'csv_export': False,  # save results as csv: False or /path/tofolder
@@ -131,7 +131,7 @@ args = {
     'extra_functionality': None,  # Choose function name or None
     # Clustering:
     'network_clustering_kmeans': 50,  # False or the value k for clustering
-    'load_cluster': 'cluster_coord_k_50_result',  # False or predefined busmap for k-means
+    'load_cluster': False, # 'cluster_coord_k_50_result',  # False or predefined busmap for k-means
     'network_clustering_ehv': False,   # clustering of HV buses to EHV buses.
     'disaggregation': None,  # None, 'mini' or 'uniform'
     'snapshot_clustering': False,  # False or the number of 'periods'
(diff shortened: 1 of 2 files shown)


etrago/tools/io.py
```diff
@@ -788,45 +790,99 @@ def decommissioning(network, session, args, **kwargs):
...
-    query = session.query(ormclass).filter(
-                        ormclass.scn_name == 'decommissioning_' + 
-                        args['scn_decommissioning'])
+            
+        if version== None:   
+            ormclass = getattr(import_module('egoio.db_tables.model_draft'),
+                           'EgoGridPfHvExtension' + comp)
+        else:
+            ormclass = getattr(import_module('egoio.db_tables.grid'),
+                           'EgoPfHvExtensionLine' + comp)
```
ulfmueller: in line 800 and 803 it seems wrong that it is '...HvExtension' and '...HvExtensionLine' ?!

ClaraBuettner: Yes, that was wrong, I changed it in my latest commit

ClaraBuettner:
With these commits, all decommissioning scenarios on oedb work well. I think the branch can be merged.

etrago/appl.py
```diff
@@ -449,10 +449,15 @@ def etrago(args):
...
-            network,
-            session,
-            args)
+        for i in range(len(args['scn_decommissioning'])):
+            network = decommissioning(
+                    network,
+                    session,
+                    version=args['gridversion'],
+                    scn_decommissioning=args['scn_decommissioning'][i],
+                    start_snapshot=args['start_snapshot'],
```
lukasoldi: Hi, here you define start_snapshot and end_snapshot for decommissioning(), but these arguments are not required by the function which leads to an error.

ClaraBuettner: Hi, thank you for your help. I fixed this in the last commit.

etrago/tools/io.py
```diff
@@ -788,45 +790,94 @@ def decommissioning(network, session, args, **kwargs):
...
+    
+    if comp =='Line':
+         comp_df = network.lines[network.lines.index.isin(df_decommisionning.index)]
+         new_cap = comp_df.s_nom-df_decommisionning.s_nom
+         network.lines.s_nom[network.lines.index.isin(df_decommisionning.index)]=\
+         new_cap
+         
+         for idx, row in network.lines.iterrows():
+             if (row['s_nom_min'] !=0) & (
+                        row['scn_name'] =='extension_' + scn_decommissioning):
```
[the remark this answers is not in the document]
ClaraBuettner: I know this error, but with commit 6246a89 I solved the problem (missing brackets). So I don't have this error anymore.

ClaraBuettner:
The error ERROR:pypsa.io:Error, new components for Carrier are not unique is caused by the extension scenarios (Carriers are added for each scenario, with the same ids and the same name). But everything works well with the error message, all carriers are there and all generators are correctly assigned, I tested this. I tried to solve the problem several times but couldn't find a solution.

lukasoldi:
Have you tested the decommissioning scenarios with a k-means clustering? While the extension seems to work fine, I receive an error when decommissioning. Have you seen something similar?

ClaraBuettner:
Yes, I tested it with k-means (and [this pypsa branch](https://github.com/openego/PyPSA/tree/feature/pypsa_13_fork)) without any errors.

ClaraBuettner:
There still seem to be some issues here so changes and testing are needed before merging. 
Removing e.g. generation capacities by decommissioning-scenarios is still a nice feature in my opinion. Maybe we leave the PR open and I will implement this later?

ulfmueller:
If we do not include this, the decommissioning function does not work, right? I think it would be nice if all args functions work for the release at least in some 'basic' way. It sounds like that you were happy with the branch but only Lukas had problems...

ClaraBuettner:
the decommissioning function works, but only for ac-lines. With these changes, all components can be decommissioned or get reduced capacities with decommissioning scenarios

ulfmueller:
ok. let's leave it out for now then.

PR #272, opened 2019-05-28
Feature/csv export iterations

ClaraBuettner:
With this branch, you can choose if saving every iteration to csv files or only the last one

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -107,8 +107,8 @@ if 'READTHEDOCS' not in os.environ:
 
 args = {
     # Setup and Configuration:
-    'db': 'oedb',  # database session
-    'gridversion': 'v0.4.5',  # None for model_draft or Version number
+    'db': 'local',  # database session
+    'gridversion': None,  # None for model_draft or Version number
     'method': 'lopf',  # lopf or pf
     'pf_post_lopf': True,  # perform a pf after a lopf simulation
     'start_snapshot': 1,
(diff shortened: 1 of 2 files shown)


ulfmueller:
Works well. But we did not think about the csv interface to eGo respectively edisgo. See [here](https://github.com/openego/eGo/blob/0edfd4e9b97da21f5600c7350d85b198a8d5d76e/ego/tools/io.py#L206-L216). The 'disaggregated' part is easy. I would suggest to name it like before. Besides that, eGo expects the final etrago network files to be in the top level folder. Here, I am not sure what is the smartest way to proceed.

ClaraBuettner:
Okay, if we keep the behavior of this branch eGo users can add '/lopf', '/pf_post_lopf' or '/disaggregated' to the variable 'csv_import_eTraGo' in eGo config.json and it works, doesn't it? 
But I agree that it's not good to change this. So we need the last results (either after pf_post_lopf or after last lopf) in the top folder. I could change the function calls so that the lopf results are only saved in 'lopf'-folder when a pf_post_lopf is calculated and the lopf or pf_post_lopf results are written into the top level folder.

ClaraBuettner:
pypsa >= 0.16.0 includes the 'ilopf'-function  (at least for the nomorepyomo-optimization) which is similar to our iterate_lopf-function but with more options (e.g. warmstarts). Maybe we can replace the iterate_lopf from etrago in further releases with the pypsa-function and don't need to update it anymore?
In addition, pypsa-ilopf can track information about the systemcosts and grid expansion in all iterations, so the desired feature is already included.

ulfmueller:
if the ego compatibility issue wouldn't be I would include as a nice feature. But since it would be a little work still to fix that it is totally ok for me to leave this out for the release.

PR #276, opened 2019-09-24
change noise to 0.01 and for the foreign generators take the max valu…

ulfmueller:
…e of the normal dist

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -426,9 +426,15 @@ def etrago(args):
         s = np.random.RandomState(args['generator_noise'])
         network.generators.marginal_cost[network.generators.bus.isin(
                 network.buses.index[network.buses.country_code == 'DE'])] += \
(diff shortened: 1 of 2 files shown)


ClaraBuettner:
I testes the branch and it worked fine, but I have some questions. Reducing the value to 0.01 causes a bigger matrix range and might lead to numerical problems, maybe we should comment that somewhere. 
Currently, the random noise isn't set for generators in the extension scenarios, changing the order could fix this. 
And the code needs much lines, so moving it in a function in utilities.py could shorten the appl.py and we could document the matrix-range-stuff there.

ClaraBuettner:
In the latest commit the noise is set in a function in utilities.py and the order of 'scn_extension' and 'random_noise' is changed.

ulfmueller:
Very nice! Thx!

PR #281, opened 2019-10-10
Update iterate_lopf

ClaraBuettner:
When using 'kirchhoff'-formulation, the model is only build in the first iteration and updated in the next ones.

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -31,7 +31,8 @@ import pypsa
 import json
 import logging
 import math
-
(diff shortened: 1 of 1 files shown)


ClaraBuettner:
pypsa >= 0.16.0 includes the 'ilopf'-function  (at least for the nomorepyomo-optimization) which is similar to our iterate_lopf-function but with more options (e.g. warmstarts). Maybe we can replace the iterate_lopf from etrago in further releases with the pypsa-function and don't need to update it anymore?

PR #283, opened 2019-10-16
kmeans bus_weighting for remove_stubs

ClaraBuettner:


ClaraBuettner:
Without this commits, it's not possible to use the same load_cluster-files with remove_stubs = True and remove_stubs = False.  It mainly changes only the order of some rows, so I think we could merge this before the next release. But since we used the remove-stubs function rarely, it's not that important.

ulfmueller:
I do not understand what it has to do with the load cluster function. Doesnt it just address the bus weighting?

etrago/cluster/networkclustering.py
```diff
@@ -517,43 +552,9 @@ def normed(x):
...
-           .groupby('bus').p_nom.sum()
-                                .reindex(network.buses.index, fill_value=0.) +
-           network.storage_units
-                                .loc[(network.storage_units.carrier
-                                      .isin(non_conv_types) == False)]
-                  .groupby('bus').p_nom.sum()
-                  .reindex(network.buses.index, fill_value=0.))
-
-    load = network.loads_t.p_set.mean().groupby(network.loads.bus).sum()
+        weight = weight.groupby(busmap.values).sum()
```
ulfmueller, line 555: why is the line now additionally necessary?

ClaraBuettner: When buses are deleted in remove_stubs clustering, there is still a weighting for the deleted stubs otherwise. In this line, the weightings of the stubs are, like the generators,  aggregated to the next remaining bus

ClaraBuettner:
> I do not understand what it has to do with the load cluster function. Doesnt it just address the bus weighting?

Sorry, I mixed up load cluster and bus_weighting files. This is needed to use the same bus_weight with and without remove_stubs function.

PR #284, opened 2019-10-25
plot OSM background 

ClaraBuettner:


ClaraBuettner:
Needs pyproj == 2.0.2 and tilemapbase == 0.4.5

ClaraBuettner:
This commit just adds the possibility to plot osm map in background of e.g. line expansion plots. I think it's not that important to add this in the next release.

ClaraBuettner:
The implementation is ok and we can add this. I'm not sure if there are any problems with the requirements of pyproj == 2.0.2 and tilemapbase == 0.4.5. I'll take a look at this.

ClaraBuettner:
I tested the installation with these new requirements and it worked. 
When we don't want to add this to the installation requirements, I could change the plotting function and only use the osm background when these packages are installed.

ulfmueller:
if the packages are easily and flawlessly installable I think it is totally ok to add it to the setup.py. so far you have not added pyproj but only tilemapbase... feature or bug? I can check it if it works for me fine as well...

ClaraBuettner:
No, it's not a feature. Due to the other installation changes I tested the installation with another local branch.

ulfmueller:
I don't understand. Doesn't pyproj has to be introduced to the setup.py?

ClaraBuettner:
Yes it has to be added to setup.py and is missing in this branch. I already tested the installation but because of the other installation issues I used a branch combined with #291

PR #285, opened 2019-10-29
Feature/snapshot clustering

ClaraBuettner:
In this branch, snapshot clustering uses p_max_pu * p_nom instead of p_set of generators for clustering.
Additionally, users can choose which constraint they would like to add for storage units in snapshot clustering, either 'daily_bounds' or 'soc_constraint'. Using 'soc_constraint' leads to better results of storage dispatch but needs more time than 'daily_bounds'.


ClaraBuettner:
Snapshot clustering is adjusted in other branches, too (e.g. https://github.com/openego/eTraGo/tree/feature/sclopf). So I think it might be easier to merge this all together at once.

ulfmueller:
anyhow, if it works fine - why not have this function in the release?!

ClaraBuettner:
it's true, we can add this to this release. Since seasonal storage constraints were added by other users too I wanted to compare the implementation first. But changes would still be possible, so we can add this now.

etrago/cluster/snapshot.py
```diff
@@ -68,30 +83,44 @@ def tsam_cluster(timeseries_df, typical_periods=10, how='daily'):
...
-    start = []
+    start=[]
     # get the first hour of the clusterCenterIndices (days start with 0)
     for i in clusterCenterIndices:
-        start.append(i * hours)
+        start.append(i*hours)
 
     # get a list with all hours belonging to the clusterCenterIndices
-    nrhours = []
+    nrhours=[]
```
ulfmueller, line 123: Were these style changes intentional?

etrago/cluster/snapshot.py
```diff
@@ -103,7 +132,42 @@ def tsam_cluster(timeseries_df, typical_periods=10, how='daily'):
...
+    for i in range(1,int(x)):
+        j=1
+        while j <= hours: 
+            nr_day.append(i)
+            j=j+1
+    df_i_h = pd.DataFrame({'Timeseries': timeseries_df.index,
+                        'Candidate_day': nr_day})
+    df_i_h.set_index('Timeseries',inplace=True)
+
+    return df_cluster, cluster_weights, dates, hours, df_i_h
```
ulfmueller, line 170: the doubling of the return statement looks very weird to me. Intentional?

etrago/cluster/snapshot.py
ClaraBuettner: No, they were not. I can't remember why I changed this, it's not pep8-conform. I can re-change that

etrago/cluster/snapshot.py
ClaraBuettner: I don't see another return statement (the one in 106 is deleted and a new one added in 170), can you help me?

ClaraBuettner:
You can choose between 'daily bounds' and 'soc_constraints' here: 
https://github.com/openego/eTraGo/blob/6470fc6b7badabce1635e4ecbe8c4da84768bdbf/etrago/appl.py#L539-L543
Maybe it's better to move it to the args, I can change this.

etrago/cluster/snapshot.py
ulfmueller: that would be nice.

ClaraBuettner: I did this in my latest commit. Looking at the old commit it only effected three lines, or did you see other issues?

etrago/cluster/snapshot.py
```diff
@@ -112,9 +176,13 @@ def run(network, n_clusters=None, how='daily',
...
 
     # calculate clusters
-    tsam_ts, cluster_weights, dates, hours = tsam_cluster(
-        prepare_pypsa_timeseries(network), typical_periods=n_clusters,
-        how=how)
+    df_cluster, cluster_weights, dates, hours, df_i_h= tsam_cluster(
+                prepare_pypsa_timeseries(network),
+                typical_periods=n_clusters,
+                how='daily',
+                extremePeriodMethod = 'None')
```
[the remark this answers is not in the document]
ClaraBuettner: I remember this error, it is solved in the feature/sclopf-branch. I'll check if I can easily copy the fix to this branch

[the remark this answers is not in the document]
ClaraBuettner: This problem is fixed in my latest commit

PR #293, opened 2020-06-12
Release 0.7.2

ulfmueller:


ClaraBuettner:
I tested the most important functions (network and snapshot clustering, pf_post_lopf, exporting) and they all worked fine. I would only change the default args['snapshot_clustering'] as described above.

ulfmueller:
>  I would only change the default args['snapshot_clustering'] as described above.

I dont know what you mean? what do you want to change to what? I dont see what upper description you refer to.

Issue #101, opened 2017-11-23
Getting rid of PyPSA fork


ulfmueller:
The dependency links to a forked repository are suboptimal. Therefore it should be checked if we can PR our changes into the mother repository and be able to use the original PyPSA.
The changes in our [dev](https://github.com/openego/PyPSA/tree/dev) and our [snapshot clustering branch](https://github.com/openego/PyPSA/tree/features/snapshot_clustering) have to be considered.

The change in the snapshot clustering branch will definitely not go into the mother repository because it displays a hard-coded individual solution. @kimvk can you check if the adjustment could be done outside of PyPSA within eTraGo?

Concerning the ehv network clustering which display the changes in the dev branch, @s3pp would you think they would be accepted by the pypsa developer or are they too specific and fit only to our problem? Should I just try a PR and if it is rejected we can try to find a solution outside of PyPSA?!

ulfmueller:
or do you disagree @BartelsJ @wolfbunke @lukasoldi ? should we just keep the fork and not bother about this stuff?
The advantage would be that concerning eGo eTraGo and eDisGo would be integrated a bit more straight foreword. Moreover we would have less problems with installation and documentation - since these fork dependencies are rather not wanted and therefore can cause problems....

lukasoldi:
I would definitely welcome an integration into the main version and make it more straightforward. But, I think we should see what @s3pp says about the issues concerning networkclustering. 
Generally, you should consider the workload for the re-integration. I remember that we decided for the fork also with the argument that we can do our own stuff independently...

wolfbunke:
I am fine with a switch to the original PyPSA. But I am not sure about the workload of the re-integration as well.

s3pp:
![20171123_pypsaforkdiff](https://user-images.githubusercontent.com/10025949/33183776-87b1f2fa-d079-11e7-91f6-098f85553dce.png)

There are only a few changes between dev and master branch, but even so I'm sure, that these changes cannot be merged back in the official PyPSA repository, because grouping on max_hours if available in aggregateoneport is a very specific change. Nevertheless, due to the fact that by now only these two functions are concerned, reimplementing those in eTraGo should not be very time-consuming.

ClaraBuettner:
PyPSA 0.17.1 allows to get rid of the fork within the next release (0.8.0). All changes of the pypsa-fork are implemented in the newer version or can be moved to etrago: 

[...]

PR #302, opened 2020-08-27
Change structure to oop and update pypsa version

ClaraBuettner:
This branch includes several changes. 
First, pypsa was updated to the newest version 0.17.1. According to this, a lot of other packages (like pandas, pyomo...) are also updated to the latest version. 
Some changes in the kmeans-clustering in etrago now allow to use this pypsa version directly, the fork is not needed anymore (described in issue #101).

In addition, the structure in etrago is changed in order to use more object-oriented programming. 
There is a new class Etrago, describing the etrago object which mainly consists of a pypsa.Network and the settings in the args. 
All functionalities, like importing data from the oedb, spatial/temporal clustering and the lopf are now part of the Etrago-object. In addition, some plotting functions and a calculation of main results can be applied directly on this object. 
This leads (at least in my mind) to a better overview of the functionalities and the appl.py which consist now of much less code. 
https://github.com/openego/eTraGo/blob/17f2a95a4bfda68fae79d95798b7bf8d5ad01d37/etrago/tools/network.py#L60-L250

I also adjusted the description in the rtd (mainly the installation instructions and the HowToUse). 
Due to installation problems on windows in the past, I thought about removing this section in the rtd or adding the advice to use a virtual machine with ubuntu. But these problems occurred because of old package version which are updated now so I suggest that windows installation will work (better) by now. I'll try to install this branch on windows and adjust the instructions by then. 

Before merging this into the dev-branch I would like to have some more opinions about it, because it includes a lot of changes which also effect the program structure.


ClaraBuettner:
When updating the rtd I recognized that the only linked example is not working anymore. (see here: https://github.com/openego/eTraGo/blob/feature/oop_aggregation/doc/howToUse.rst#examples-and-tutorial-notebooks) 
I think it would be better to delete this section (since the default settings in the appl.py could be used as an example) or add new running examples that fit to the new version of etrago.

wheitkoetter (approved):
As I just started as an etrago developer, I am not able to assess all the changes in-depth. However, I did not find any problems and approve the changes.

viole-i:
I like that we won't need the dependency links anymore.

viole-i:
Like Wilko, I am quite new to eTraGo. I could not find any bugs while installing the new version or running it. Approved from my side.

viole-i:
Accidentally closed, sorry.

viole-i (approved):

ClaraBuettner:
I just tried to install this eTraGo branch on one of our servers in Flensburg which still uses python 3.5. The installation works but running etrago throws errors because pypsa is only compatible to python 3.6 or 3.7 since version 0.16.0 (e.g. f-strings are used which were added to the python syntax in version 3.6). In general, it is possible to change all these functions, but be would need a pypsa-fork and a lot of time for the adjustments. 
In addition, when using python 3.5 I recently get warnings:
 `DEPRECATION: Python 3.5 reached the end of its life on September 13th, 2020. Please upgrade your Python as Python 3.5 is no longer maintained. pip 21.0 will drop support for Python 3.5 in January 2021. pip 21.0 will remove support for this functionality.`
I think it might be the best option to loose the support for python 3.5 like pypsa. But that means that we need to update our servers. Which version do you use @wheitkoetter @viole-i ?

ulfmueller (approved):

PR #306, opened 2020-10-05
Add line length factor to kmeans clustering

ClaraBuettner:
The line_length_factor was not implemented correctly in kmeans clustering. This bug is fixed in the following commit.

### DIFF
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -628,7 +628,8 @@ def kmean_clustering(etrago):
                               'p_nom_opt': np.sum,
                               'marginal_cost': np.mean,
                               'capital_cost': np.mean},
-        aggregate_one_ports=aggregate_one_ports)
+        aggregate_one_ports=aggregate_one_ports,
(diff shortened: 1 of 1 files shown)


ulfmueller (dismissed):

ulfmueller:
the line length factor is only set in the 'remove_stubs' if section. In my mind, it should be also part of the function at this [place](https://github.com/openego/eTraGo/blob/7161c858c02704963b897bbb1faffdf4622638ae/etrago/cluster/networkclustering.py#L616) Do you agree?

ClaraBuettner:
You describe the old bug, which sholud be fixed in this branch. I can't see to which branch/commit your link belongs but in this branch, it is fixed for remove_stubs and kmeans: 
https://github.com/openego/eTraGo/blob/e83b3a7e78af9a01016bf34096e0fb0748277fff/etrago/cluster/networkclustering.py#L616-L632

ulfmueller:
you are so right. sorry, my bad - monday groove... :wink:

PR #319, opened 2021-02-21
bugfix - add .abs() in plot.py, calc_ac_loading function

wheitkoetter:
Hi Clara and Ulf,

I hope you are doing well :)

Does there not have to be such an .abs() to be added as proposed in the commit? Because the average shall be calculated over both negative and positive flows. In line 1158 at the calculation for q0 there is also such an .abs() included: mul_weighting(network, network.lines_t.q0).loc[network.snapshots[timesteps]].abs().sum()

Or am I wrong?

Cheers Wilko

### DIFF
--- a/etrago/tools/plot.py
+++ b/etrago/tools/plot.py
@@ -1149,7 +1149,7 @@ def calc_ac_loading(network, timesteps):
     """
 
     loading_lines = mul_weighting(network, network.lines_t.p0).loc[
-                network.snapshots[timesteps]].sum()
+                network.snapshots[timesteps]].abs().sum()
 
     if not network.lines_t.q0.empty:
 


ClaraBuettner:
Hi Wilko, 
sorry for my late reply. I hope you are also doing well!
I think both options are reasonable. I often used this function to visualize transfer flows between different buses, therefore I had to sum up negative and positive values. When you want to see the mean loading of the line e.g. to see how often it is used, the absolute values have to be summed up. But it is definitely wrong to do it differently for active and reactive power. 
I would suggest to have both options; 'line_loading' (with `.abs()`) and 'line_flows' (without `abs()`). The arrows on the lines would be used only in the 'line_flows' plot.

PR #331, opened 2021-04-07
change some setup requirements

ulfmueller:


ClaraBuettner (approved):
You have also changed the default args, which is totally fine (it might be also better for testing because it is much faster). I tested the installation and run with the default args with python3.6 and 3.7.

PR #339, opened 2021-10-14
Features/#322 snapshot clustering constraints without pyomo

ClaraBuettner:
This branch includes the snapshot clustering constraints without pyomo. It seems to work fine, but I only had a brief look at the results.
 @KathiEsterl Could you maybe check more in detailed if it works as expected? I can explain you how you can extract the values for `soc_intra` from the results.

### DIFF
--- a/etrago/tools/constraints.py
+++ b/etrago/tools/constraints.py
@@ -1149,6 +1149,8 @@ def snapshot_clustering_seasonal_storage(self, network, snapshots):
             p_nom = m.storage_p_nom[s]
         else:
             p_nom = network.storage_units.p_nom[s]
+            
+        import pdb; pdb.set_trace()
 
         return (m.state_of_charge_intra[s, intra_hour] +
                 m.state_of_charge_inter[s, network.cluster_ts['Candidate_day'][h]]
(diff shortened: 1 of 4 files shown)


etrago/tools/constraints.py
```diff
@@ -1231,33 +1244,18 @@ def snapshot_clustering_seasonal_storage_nmp(self, n, sns):
...
-    last_soc_total = soc_total.loc[last_hour].set_index(candidates)
-    last_soc_inter = soc_inter.shift(1).fillna(soc_inter.loc[candidates[-1]])
+    last_soc_intra = soc_intra.loc[last_hour].set_index(candidates)
 
     coeff_var = [(-1, next_soc_inter),
                  (eff_stand.pow(24), soc_inter),
-                 (eff_stand, last_soc_total),
-                 (-eff_stand, last_soc_inter),
+                 (eff_stand, last_soc_intra),
                  (-1/eff_dispatch, dispatch),
```
KathiEsterl, line 1252: Do we need to change p_dispatch and p_store to the values of the next hour which is the first hour of the next representative day? The soc_intra is defined that way in this code. Is this just a definition question or do we have to align this constraint to avoid mixed definitions?

KathiEsterl:
I changed the constraints for the soc_inter now using the power flow of the first hour of the next representative day to calculate the needed soc_intra. I checked all the constraints manually in some test calculations again - everything should be fine now!

KathiEsterl:
> I changed the constraints for the soc_inter now using the power flow of the first hour of the next representative day to calculate the needed soc_intra. I checked all the constraints manually in some test calculations again - everything should be fine now!

One thing to add: I left **efficiencies = 1** in the appl.py in case you also want to check it one more time. The rest should be okay.

KathiEsterl:
I checked the last adaptions, deleted the unnecessary parts and checked it one more time manually :+1:

KathiEsterl (approved):

PR #341, opened 2021-10-29
Features/#333 spatial clustering for sector coupling

CarlosEpia:
The applied changes allow to cluster all the different sectors around the eHV grid.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -327,7 +327,7 @@ def run_etrago(args, json_path):
         eTraGo result network based on `PyPSA network
         <https://www.pypsa.org/doc/components.html#network>`_
     """
-    etrago = Etrago(args, json_path= None)
+    etrago = Etrago(args, json_path)
  
     # import network from database
     etrago.build_network_from_db()
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -90,7 +90,7 @@ args = {
         'max_iter': 100, # affects clustering algorithm, only change when neccesary
         'tol': 1e-6, # affects clustering algorithm, only change when neccesary
         'n_jobs': -1}, # affects clustering algorithm, only change when neccesary
-    'network_clustering_ehv': False,  # clustering of HV buses to EHV buses.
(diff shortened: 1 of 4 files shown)


ClaraBuettner:
I finally found the problem with the empty `etrago.network.loads_t.p_set`; I set `with_time ` to `False `here:
https://github.com/openego/eTraGo/blob/630a2189a2003eab4c2e02fb17d4331df6a76ebd/etrago/cluster/networkclustering.py#L424-L427
I think I did this  because there were some problems when no load data was available. But now there is load data available, and to consider it in clustering this parameter has to be set to `True` again. Sorry for the confusion! I fixed this in my latest commit.

etrago/appl.py
```diff
@@ -327,8 +327,8 @@ def run_etrago(args, json_path):
         eTraGo result network based on `PyPSA network
         <https://www.pypsa.org/doc/components.html#network>`_
     """
-    etrago = Etrago(args, json_path)
-
+    etrago = Etrago(args, json_path= None)
```
ClaraBuettner, line 330: Do you want to merge this? It looks more like a temporal adjustment...

etrago/cluster/networkclustering.py
```diff
@@ -125,8 +217,8 @@ def cluster_on_extra_high_voltage(network, busmap, with_time=True):
     network.generators.control = "PV"
     network.generators['weight'] = 1
 
-    network.generators = network.generators[~network.generators.carrier.isin(
-        ['gas', 'solar_thermal_collector', 'geo_thermal'])]
+    #network.generators = network.generators[~network.generators.carrier.isin(
+    #    ['gas', 'solar_thermal_collector', 'geo_thermal'])]
```
ClaraBuettner, line 221: These lines could be deleted

etrago/appl.py
CarlosEpia: The "= None" was deleted.

etrago/cluster/networkclustering.py
CarlosEpia: The mentioned lines were deleted.

PR #337, opened 2021-10-04
Kmedoids dijkstra clustering

KathiEsterl:
Implementation for k-medoids Dijkstra Clustering; added interim solutions for spatial clustering

see #336

PR #351, opened 2022-01-02
Features/#333 spatial clustering for sector coupling

CarlosEpia:
This branch includes the eHV and k-mean clustering adjusted for the new network including the new sectors.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,12 +17,12 @@ energy storages. Grid expansion measures are not part of this tool and will be
 instead part of 'eGo' https://github.com/openego/eGo
 
 .. warning::
-  From now on eTraGo depends on a sector coupled data-model. This is not published on 
-  the oedb yet, the data has to be created using
+  From now on eTraGo depends on a sector coupled data-model created by
   `eGon-data <https://github.com/openego/eGon-data>`_.
   Not all functions and features are compatible to the sector coupled model yet.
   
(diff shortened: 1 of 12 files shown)


ulfmueller:
I started testing it but so far I run in an error which is called inside of the [pypsa aggregatelines function](https://github.com/PyPSA/PyPSA/blob/v0.17.1/pypsa/networkclustering.py#L204) or more precisely in the consense function and assertion error like the following occurs:


  File "/home/ulf/venvs/etrago/lib/python3.7/site-packages/pypsa/networkclustering.py", line 56, in consense
    .format(component, x.name, attr, x)

AssertionError: In Bus cluster type the values of attribute type do not agree:
27613                               
27625                               
27626                               
27627                               
27628                               
27629                               
27630                               
27639                               
27567    Al/St 240/40 4-bundle 380.0
Name: type, dtype: object

Any ideas? What dataset are you using?

ulfmueller:
I think it makes sense to pull from dev at this point. I tried it but considering your work in networkclustering.py I think @CarlosEpia , you should do it, so I stopped and aborted the merge. ok?

CarlosEpia:
This branch is able to perform the eHV and k-mean clustering, but the k-mean depends on the eHV. It is something I will change with issue #357. Once it is done, the PR #337 can be merged.

PR #355, opened 2022-01-18
Add constraints for CHP links

### LINKED ISSUE #345: Integrate constraints for CHP plants
CHPs are modeled as two link components, a `CHP` from a gas bus to an electricity bus and a `CHP heat` from a gas to a heat bus. The dispatch of these links will be connected by constraints.

ClaraBuettner:
Closes #345


fwitte:
Just FYI: We discovered, that after clustering the number of links are not identical anymore. See the attached file, which was set up with one CH4 bus only but 33 electricity buses in DE for clustering.

With the default strategy, the clustering for heat is first based on CH4 grid, then on AC grid. After that, the central_gas_CHP_heat link corresponds to the 33 central_gas_CHP links, which will crash the constraint building. If I find a possible solution, I'll let you know! 

[links.csv](https://github.com/openego/eTraGo/files/8823961/links.csv)

Issue #343, opened 2021-11-08
k-mean clustering for the gas sector


AmeliaNadal:
The first implementation of k-means clustering for the new sectors will be done similar to the ehv-clustering adjustments. First, the electrical grid will be clustered and the new sectors will be added accordingly afterwards.

Issue #358, opened 2022-02-03
Standardize the gas clustering 


AmeliaNadal:
The first implementation of the gas clustering available in the branch [features/%23343_with_333](https://github.com/openego/eTraGo/tree/features/%23343_with_333) is working only for the eGon2035 scenario datasets. It should be more generic.

PR #359, opened 2022-02-03
First implementation of gas clustering

AmeliaNadal:
This branch is suppose to allow the clustering of the gas grid : CH4, H2_grid and H2_ind_load buses, related links, generators, loads and stores.

Important remarks:
- The gas clustering should append after the clustering of the other new sectors or more precisely after the clustering of the H2_saltcaverns buses. Therefore, the branches [features/356_assign_country_to_buses ](https://github.com/openego/eTraGo/tree/features/356_assign_country_to_buses) and [features/#333_spatial_clustering_for_sector_coupling](https://github.com/openego/eTraGo/tree/features/%23333_spatial_clustering_for_sector_coupling) where merged into this branch.
- This branch has been tested only on a reduce dataset (test mode of egon-data) and without the CHP technologies.
- This clustering is functionning only for the eGon2035 scenario datasets, this should become more generic: see https://github.com/openego/eTraGo/issues/358


etrago/cluster/gasclustering.py
```diff
@@ -0,0 +1,418 @@
...
+        n_clusters=kmean_gas_settings["n_clusters_gas"],
+        n_init=kmean_gas_settings["n_init"],
+        max_iter=kmean_gas_settings["max_iter"],
+        tol=kmean_gas_settings["tol"],
+        n_jobs=kmean_gas_settings["n_jobs"],
+    )
+
+    # Add H2_grid buses to busmap
+    engine = db.connection(section=etrago.args["db"])
+    sql = f"""SELECT "bus_H2", "bus_CH4", scn_name FROM grid.egon_etrago_ch4_h2;"""
```
ClaraBuettner, line 115: Other tables are imported in etrago.io. If this table is needed we have to make sure that it is part of the eTraGo input database (and e.g. also publish it).

etrago/cluster/gasclustering.py
```diff
@@ -0,0 +1,418 @@
...
+        etrago.network,
+        gas_busmap,
+        bus_strategies={
+            "country": "first",
+        },
+        one_port_strategies={
+            "Generator": {
+                "marginal_cost": np.mean,
+                "capital_cost": np.mean,
+                "p_nom_max": np.max,
```
ClaraBuettner, line 374: I think you should use` np.sum` for `p_nom_max` and `p_nom_min` because you want to sum up the capacity limits.

etrago/appl.py
```diff
@@ -85,7 +86,7 @@
         'max_iter': 100, # affects clustering algorithm, only change when neccesary
         'tol': 1e-6, # affects clustering algorithm, only change when neccesary
         'n_jobs': -1}, # affects clustering algorithm, only change when neccesary
```
ClaraBuettner, line 88: In a new environment, the clustering is not working because the `n_jobs` parameter was removed in scikit learn in  versions `>0.23`. We can either restrict the package version or remove the parameter from our functions. 
I have the tendency to use the second option to avoid new conflicts of package versions. What do you think?
@CarlosEpia: It is the same for clustering the electricity grid, we should agree one one solution

CarlosEpia: I think the best option is to remove the n_jobs parameter. I will do it in my branch #361

etrago/cluster/gasclustering.py
AmeliaNadal: Thanks for the correction, I've fixed this in the the commit https://github.com/openego/eTraGo/pull/359/commits/8b7ea56190e2d833fa9e28783c3c8cd29a0a5057.

ClaraBuettner:
> This branch is suppose to allow the clustering of the gas grid : CH4, H2_grid and H2_ind_load buses, related links, generators, loads and stores.
> 
> Important remarks:
> 
>     * The gas clustering should append after the clustering of the other new sectors or more precisely after the clustering of the H2_saltcaverns buses. Therefore, the branches [features/356_assign_country_to_buses ](https://github.com/openego/eTraGo/tree/features/356_assign_country_to_buses) and [features/#333_spatial_clustering_for_sector_coupling](https://github.com/openego/eTraGo/tree/features/%23333_spatial_clustering_for_sector_coupling) where merged into this branch.
> 
>     * This branch has been tested only on a reduce dataset (test mode of egon-data) and without the CHP technologies.
> 

I tested the gas clustering for Schleswig-Holstein and Germany and it was working as expected. The CHPs are also clustered correctly.

CarlosEpia:
I tested this branch and it is running for me. It looks ok. I have two questions about the methodology:
1. This approach clusters each kind of gas-bus according to the parameter 'n_clusters_gas'. Is that actually what we want?
2. The clustering is not considering the buses in other countries, which means that in the end, we have more buses than the given by the 'n_clusters_gas' parameter. It works differently for the eHV and k-mean clustering for the other sectors. Should we agree on a unique approach regarding this topic?

etrago/appl.py
```diff
@@ -74,7 +74,8 @@
     # Clustering:
     'network_clustering_kmeans': {
         'active': True, # choose if clustering is activated
-        'n_clusters': 10, # number of resulting nodes
+        'n_clusters': 20, # number of resulting nodes
+        'n_clusters_gas': 5, # number of resulting nodes
```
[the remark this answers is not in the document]
AmeliaNadal: This issue will be discussed further here: https://github.com/openego/eTraGo/issues/364.

etrago/cluster/gasclustering.py
ClaraBuettner: I think the new option is better, thank you!

etrago/cluster/gasclustering.py
```diff
@@ -0,0 +1,397 @@
...
+        lambda x: tuple(sorted([str(x.bus0), str(x.bus1)])), axis=1
+    )
+
+    strategies.update(
+        {col: "first" for col in combinations.columns if col not in strategies}
+    )
+
+    combinations_final = combinations.groupby(["buscombination", "carrier"]).agg(
+        strategies
+    )
```
ClaraBuettner, line 299: I'm wondering why you are clustering the links twice (line 286 and line 297). Do you want to cluster links bidirectional?

etrago/cluster/gasclustering.py
```diff
@@ -0,0 +1,397 @@
...
+        new_links[~new_links["carrier"].isin(gas_carriers)]
+        .copy()
+        .drop(columns="link_id")
+    )
+    io.import_components_from_dataframe(network_gasgrid_c, non_gas_links, "Link")
+
+    if with_time:
+        for attr, df in iteritems(network.links_t):
+            if not df.empty:
+                io.import_series_from_dataframe(network_gasgrid_c, df, "Link", attr)
```
ClaraBuettner, line 315: Time-depended values for links are imported, but there are time series values for some links_id which are not in the links dataframe :
```

WARNING:pypsa.io:Components Index(['1740', '1741', '1742', '1743', '1744', '1745', '1746', '1747', '1748',
       '1749',
       ...
       '2005', '2006', '2007', '2008', '2009', '2010', '2011', '2012', '2013',
       '2014'],
      dtype='object', length=216) for attribute efficiency of Link are not in main components dataframe links
```
I will try to find out why this happens.

ClaraBuettner: The warnings seem to be related to #366. It could be solved by changing the order of the clustering and clustering the gas network before clustering the electric grid. Is there anything against changing the execution order like this?

CarlosEpia: I also think it is related to #366. The clustering order must be irrelevant, so if it helps to swap the clusterings, I am not against it.

fwitte: I am working on https://github.com/openego/eTraGo/pull/367, it should be possible to modify the aggragation that in a way that the "in-between" technologies (generalized: carriers that are connected to clustered network structures) are aggregated after the busmaps for gas and AC have been created, indendetly of the clustering order.

etrago/cluster/gasclustering.py
ClaraBuettner: But there could be problems when you have unidirectional links which should be not combined with links going into the other direction. If you have in the network with the full complexity unidirectional links in both directories, you don't want to combine them and have only one link into one direction. Or am I mistaken?

etrago/cluster/gasclustering.py
ClaraBuettner: So how do we want to proceed? Should we wait until #367 is finished and merged into this branch before merging it into dev?

CarlosEpia: For me, it makes sense to merge #367 into this branch first and then check that everything is working together as @ClaraBuettner suggested.

ClaraBuettner: This issue will be solved after #377 is merged into dev.

ClaraBuettner (approved):

AmeliaNadal:
Closes https://github.com/openego/eTraGo/issues/343

Issue #356, opened 2022-01-19
Country column in buses df is missed after clustering


CarlosEpia:
Currently, each bus has an assigned country in the egon-data db. When they are imported and clustered using the pypsa function  "aggregatebuses", the column "country" is lost. Using the function "buses_by_country" becomes necessary, but it is not working properly  (some buses are assigned to more than one country). As I see it, we can improve the used boundaries or change the methodology to use Geopandas and the shape of the neighboring countries.

ClaraBuettner:
> Using the function "buses_by_country" becomes necessary, but it is not working properly (some buses are assigned to more than one country). As I see it, we can improve the used boundaries or change the methodology to use Geopandas and the shape of the neighboring countries.

Before I committed 30b75f3b4accbb49b102dc5ae68eee073301c631, your second option was also an option. But by now, we don't included the shape of neighboring countries into the data base. 
When we need it in eTraGo, we can add it to the database or try to find another way (e.g. a tool) to get the needed data.

CarlosEpia:
The shape of all the countries in the world is included by default in Geopandas. You can read it by typing:

path = gpd.datasets.get_path("naturalearth_lowres")
boundaries = gpd.read_file(path)

I would have to check if the resolution is high enough to perform this task. Do you think that we can use this shapefile to assign the country data to our buses?

ClaraBuettner:
> The shape of all the countries in the world is included by default in Geopandas. You can read it by typing:
> 
> path = gpd.datasets.get_path("naturalearth_lowres") boundaries = gpd.read_file(path)
> 
> I would have to check if the resolution is high enough to perform this task. Do you think that we can use this shapefile to assign the country data to our buses?

Thank you, I didn't know about this feature... If the resolution is fine, we can use it.

PR #360, opened 2022-02-04
Features/356 assign country to buses

CarlosEpia:
Fixes issue #356. Geopandas must be installed using PIP.


ClaraBuettner:
It seems like this branch also includes some changes for the clustering, or am I mistaken? I'm just wondering in which order we should merge the feature branches...

ClaraBuettner (approved):

PR #367, opened 2022-02-09
Clustering of sector coupled carriers

fwitte:
- Add a method that finds the last bus_id available
- Add a method that clusters buses which a connected to already clustered buses, but have not been clustered before (e. g. heat buses connected to AC and gas)

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -88,18 +88,10 @@ args = {
         'tol': 1e-6,}, # affects clustering algorithm, only change when neccesary
     'sector_coupled_clustering': {
         'active': True, # choose if clustering is activated
-        'carrier_data': {
-            'H2_ind_load': { # key: name of the carrier for the buses to cluster
+        'carrier_data': { # select carriers affected by sector coupling
+            'H2_ind_load': {
                 'base': ['H2_grid'],
-                # list of carriers, that clustering should be topologically
-                # based on
                 'skip': None,
-                # list of carriers, that are skipped in topological bus
-                # clustering, but will be aggregated on the clustered buses in
-                # a second step. E.g. every central_heat bus has a respective
-                # central_heat_store bus. central_heat will be clustered as
-                # provided in the config, and the central_heat_store will be
-                # aggregated on the clustered central_heat buses afterwards
             },
             'central_heat': {
                 'base': ['CH4'],
@@ -109,7 +101,7 @@ args = {
                 'base': ['CH4'],
                 'skip': 'rural_heat_store'
             },
-        }, # add necessary information for sector coupled carriers
+        },
     },
     'network_clustering_ehv': False,  # clustering of HV buses to EHV buses.
     'disaggregation': 'uniform',  # None, 'mini' or 'uniform'
@@ -307,6 +299,21 @@ def run_etrago(args, json_path):
         changed carefully, documentation and possible settings are described
         in sklearn-package (sklearn/cluster/k_means_.py).
         This function doesn't work together with ``'line_grouping = True'``.
+
+    sector_coupled_clustering : nested dict
+        {'active': True, 'carrier_data': {
+         'H2_ind_load': {'base': ['H2_grid'], 'skip': None, },
+         'central_heat': {'base': ['CH4'], 'skip': 'central_heat_store'},
+         'rural_heat': {'base': ['CH4'], 'skip': 'rural_heat_store'}}
+        }
+        State if you want to apply clustering of sector coupled carriers, such
+        as central_heat or rural_heat. The approach builds on already clustered
+        buses (e.g. CH4 and AC) and builds clusters around the topology of the
+        buses with carrier ``'base'`` for all buses of a specific carrier, e.g.
+        ``'H2_ind_load'``. Connections (via links) of the selected buses to
+        buses of the carrier specified in ``'skip'`` are ignored in the
+        clustering and aggregated based on the newly clustered buses in a
+        following step.
 
     network_clustering_ehv : bool
         False,
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -89,12 +89,20 @@ args = {
     'sector_coupled_clustering': {
         'active': True, # choose if clustering is activated
         'carrier_data': {
-            'H2_ind_load': {
-                'base': ['H2_grid', 'H2_saltcavern'],
+            'H2_ind_load': { # key: name of the carrier for the buses to cluster
+                'base': ['H2_grid'],
+                # list of carriers, that clustering should be topologically
+                # based on
                 'skip': None,
+                # list of carriers, that are skipped in topological bus
+                # clustering, but will be aggregated on the clustered buses in
+                # a second step. E.g. every central_heat bus has a respective
+                # central_heat_store bus. central_heat will be clustered as
+                # provided in the config, and the central_heat_store will be
+                # aggregated on the clustered central_heat buses afterwards
             },
             'central_heat': {
-                'base': ['AC', 'CH4'],
+                'base': ['CH4'],
                 'skip': 'central_heat_store'
             },
             'rural_heat': {
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -92,15 +92,12 @@ args = {
         'carrier_data': { # select carriers affected by sector coupling
             'H2_ind_load': {
                 'base': ['H2_grid'],
-                'skip': None,
             },
             'central_heat': {
                 'base': ['CH4'],
-                'skip': 'central_heat_store'
             },
             'rural_heat': {
                 'base': ['CH4'],
-                'skip': 'rural_heat_store'
             },
         },
     },
@@ -305,18 +302,15 @@ def run_etrago(args, json_path):
 
     sector_coupled_clustering : nested dict
         {'active': True, 'carrier_data': {
-         'H2_ind_load': {'base': ['H2_grid'], 'skip': None, },
-         'central_heat': {'base': ['CH4'], 'skip': 'central_heat_store'},
-         'rural_heat': {'base': ['CH4'], 'skip': 'rural_heat_store'}}
+         'H2_ind_load': {'base': ['H2_grid']},
+         'central_heat': {'base': ['CH4']},
+         'rural_heat': {'base': ['CH4']}}
         }
         State if you want to apply clustering of sector coupled carriers, such
         as central_heat or rural_heat. The approach builds on already clustered
         buses (e.g. CH4 and AC) and builds clusters around the topology of the
         buses with carrier ``'base'`` for all buses of a specific carrier, e.g.
-        ``'H2_ind_load'``. Connections (via links) of the selected buses to
-        buses of the carrier specified in ``'skip'`` are ignored in the
-        clustering and aggregated based on the newly clustered buses in a
-        following step.
+        ``'H2_ind_load'``.
 
     network_clustering_ehv : bool
         False,
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -371,8 +371,6 @@ def run_etrago(args, json_path):
     # k-mean clustering
     etrago.kmean_clustering()
     etrago.kmean_clustering_gas()
-    s
-
 
     etrago.args['load_shedding']=True
     etrago.load_shedding()
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -343,13 +343,13 @@ def run_etrago(args, json_path):
     etrago.network.transformers.tap_ratio.fillna(1, inplace=True)
     etrago.network.buses.v_mag_pu_set.fillna(1., inplace=True)
     etrago.network.loads.sign = -1
-    
+   
     # ehv network clustering
-    # etrago.ehv_clustering()
+    etrago.ehv_clustering()
 
     # k-mean clustering
     etrago.kmean_clustering()
-    
+
     # k-mean clustering for gas and related sectors
     etrago.kmean_clustering_gas()
 
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -86,6 +86,23 @@ args = {
         'n_init': 10, # affects clustering algorithm, only change when neccesary
         'max_iter': 100, # affects clustering algorithm, only change when neccesary
         'tol': 1e-6,}, # affects clustering algorithm, only change when neccesary
+    'sector_coupled_clustering': {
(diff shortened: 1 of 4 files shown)


fwitte:
To make the draft fully working,

- a function will be applied to the network after creating the gas side busmap in the gas_clustering, that aggregates the sector coupling technologies.
- a flag will be added, to decide which version of aggregation in the sector coupling will be applied, something like
  - `preserve_topology`: only clusters buses, which are connected to the same sector clusters with the same links (most complex variant)
  - `carrier_centered`: clusters the buses around the chosen carrier. A clustered bus will retain all links to carriers different from the carrier specified in this keyword.
  - Both variants will need some more information, e.g. on how to deal with buses, that are only connected to the buses that should be clustered.

fwitte:
@CarlosEpia: If you want, you can try out the clustering for sector coupling.

You can modfiy the approach by using one of the lines here:

https://github.com/openego/eTraGo/blob/2d8ac71e31744231c7f702203d36a9fdf41c08f9/etrago/cluster/gasclustering.py#L138-L140

Currently the more complex version is active, if you comment that out and use the "CH4" version instead, the simplified version will be applied. I will add some kind of wrapper to make it possible for users to control this behavior next week.

@ClaraBuettner: It seems, some heat buses are connected to multiple gas buses via CHP if I am not mistaken. We could have a look at this together in the next week.

CarlosEpia:
@fwitte I just checked the changes and they look good for me. Since you want to merge #358 into a branch that is not dev, I would suggest merging them and check how they work once they are integrated. Does it make sense for you?

fwitte:
> Since you want to merge #358 into a branch that is not dev, I would suggest merging them and check how they work once they are integrated.

Good point, I changed the base to dev. I merged dev into this branch anyway. I will go ahead and create some kind of functionality control along with documentation and aim to merge by the end of the week.

fwitte:
The lines 

https://github.com/openego/eTraGo/blob/9310a9f31bd2f1d0fdd9c69dcec7edcb9ae3ad02/etrago/appl.py#L89-L105

let the user control the specifics in sector coupling. The example is tested with the test mode dump from 07.02., note that H2_ind_load buses are not included in that dump. Additional tests should be carried out, if we have new data.

ClaraBuettner:
> The lines
> 
> https://github.com/openego/eTraGo/blob/9310a9f31bd2f1d0fdd9c69dcec7edcb9ae3ad02/etrago/appl.py#L89-L105
> 

I like the idea to set these values in the arguments. But without having a closer look into the code I'm a bit wondering about the default parameters. Could you explain to me why `central_heat` is clustered based on CH4 and AC and `rural_heat` clustering is only based on CH4? There are P2H technologies for both carriers, so I don't really understand why they are clustered differently. 

> let the user control the specifics in sector coupling. The example is tested with the test mode dump from 07.02., note that H2_ind_load buses are not included in that dump. Additional tests should be carried out, if we have new data.

A new dump from the recent CI-run is uploaded in the Nextcloud folder (if you don't have the link I can send it to you via mail, since its public available I don't want to share it on GitHub).

fwitte:
> I like the idea to set these values in the arguments. But without having a closer look into the code I'm a bit wondering about the default parameters. Could you explain to me why `central_heat` is clustered based on CH4 and AC and `rural_heat` clustering is only based on CH4? There are P2H technologies for both carriers, so I don't really understand why they are clustered differently.

Well, the intention was to show, how the parameters can be adjusted. I'll just add a template at the very beginning, that does nothing instead.

> I couldn't find such CHPs in my database. Do you have an example where this happens?

In that dump specifically: After AC clustering and before gas clustering (and sector coupling) the buses 5990, 5985, 5974, 5978 (central_heat) are connected via central_gas_CHP_heat to buses 5173 and 5174, 5174 and 5182, 5183 and 5184, 5183 and 5184 (CH4).

fwitte:
grafik

grafik

fwitte:
CHP carriers issue resolve, data is correct. Both clustering approaches presetend work on the test mode for the heat sector.

fwitte:
When trying to run the clustering on the latest CI full Germany run, I do get an error in the k-mean-clustering, can we have a look together @CarlosEpia?

```
Traceback (most recent call last):
  File "etrago/appl.py", line 438, in <module>
    etrago = run_etrago(args, json_path=None)
  File "etrago/appl.py", line 404, in run_etrago
    etrago.kmean_clustering()
  File "/home/witt_fa/eGon/eTraGo/etrago/cluster/networkclustering.py", line 984, in run_kmeans_clustering
    self.clustering = kmean_clustering(self)
  File "/home/witt_fa/eGon/eTraGo/etrago/cluster/networkclustering.py", line 906, in kmean_clustering
    network, busmap = adjust_no_electric_network(network, busmap, cluster_met="k-mean")
  File "/home/witt_fa/eGon/eTraGo/etrago/cluster/networkclustering.py", line 118, in adjust_no_electric_network
    & (network2.links["carrier"] == map_carrier[carry])
KeyError: 'dsm'
```

CarlosEpia:
There were 3 kinds of DSM carriers that were merged in https://github.com/openego/eGon-data/tree/features/%23661-aggregate-dsm-components. I will update the clustering according to this change tomorrow.

CarlosEpia:
Hi @fwitte, is it ok for you if I push directly in this branch the changes to solve the problem you are having with the DSM buses?

fwitte:
Of course!

CarlosEpia:
It is done. Can you give it a try to confirm that it is working for you?

fwitte:
@CarlosEpia, it does work on testmode. Results for heat sector and H2_ind_gas look fine. In my opinion we can merge this.

However, there are new issues regarding electricity network clustering. I do get the following error, but it looks like this is an issue emerging from the data and is not related to the changes made here in this PR.

```
Traceback (most recent call last):
  File "etrago/appl.py", line 438, in <module>
    etrago = run_etrago(args, json_path=None)
  File "etrago/appl.py", line 404, in run_etrago
    etrago.kmean_clustering()
  File "/home/witt_fa/eGon/eTraGo/etrago/cluster/networkclustering.py", line 985, in run_kmeans_clustering
    self.clustering = kmean_clustering(self)
  File "/home/witt_fa/eGon/eTraGo/etrago/cluster/networkclustering.py", line 913, in kmean_clustering
    clustering = get_clustering_from_busmap(
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 255, in get_clustering_from_busmap
    buses, linemap, linemap_p, linemap_n, lines, lines_t = get_buses_linemap_and_lines(network, busmap, line_length_factor, bus_strategies, with_time)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 235, in get_buses_linemap_and_lines
    lines, linemap_p, linemap_n, linemap, lines_t = aggregatelines(network, buses, interlines, line_length_factor, with_time)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 203, in aggregatelines
    lines = interlines_c.groupby(['bus0_s', 'bus1_s']).apply(aggregatelinegroup)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pandas/core/groupby/groupby.py", line 1275, in apply
    result = self._python_apply_general(f, self._selected_obj)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pandas/core/groupby/groupby.py", line 1309, in _python_apply_general
    keys, values, mutated = self.grouper.apply(f, data, self.axis)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pandas/core/groupby/ops.py", line 852, in apply
    res = f(group)
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 200, in aggregatelinegroup
    data.update((f, consense[f](l[f])) for f in columns.difference(data))
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 200, in <genexpr>
    data.update((f, consense[f](l[f])) for f in columns.difference(data))
  File "/home/witt_fa/anaconda3/envs/etrago_env/lib/python3.8/site-packages/pypsa/networkclustering.py", line 49, in consense
    assert ((x == v).all() or x.isnull().all()), (
AssertionError: In Bus cluster lifetime the values of attribute lifetime do not agree:
Line
27566     inf
27627    40.0
27628    40.0
27631    40.0
27632    40.0
Name: lifetime, dtype: float64
```

It would be great, if you both review @CarlosEpia and @ClaraBuettner :)

etrago/appl.py
```diff
@@ -87,9 +87,26 @@
...
+            'central_heat': {
+                'base': ['CH4'],
+                'skip': 'central_heat_store'
+            },
+            'rural_heat': {
+                'base': ['CH4'],
+                'skip': 'rural_heat_store'
+            },
+        },
+    },
```
CarlosEpia, line 106: What do you think about setting these parameters as default (inside networkclustering.py)?

CarlosEpia (commented):
I ran it successfully with the test case and the results look fine. In my opinion, it can be merged into dev, but I would wait for @ulfmueller opinion.

etrago/appl.py
fwitte: Why inside networkclustering and not appl.py?

ulfmueller:
@CarlosEpia do you want me to review the whole PR or just get an opinion on s.th. in particular?

CarlosEpia:
@ulfmueller it is the last modification we are planning to do in the clustering methods (besides #337) and Dev was recently merged into this branch. I already ran it and checked the results, and they look good to me. Could you run it and check if you find something that I may have missed in the results after the optimization?

etrago/appl.py
```diff
@@ -286,25 +303,40 @@ def run_etrago(args, json_path):
...
+        }
+        State if you want to apply clustering of sector coupled carriers, such
+        as central_heat or rural_heat. The approach builds on already clustered
+        buses (e.g. CH4 and AC) and builds clusters around the topology of the
+        buses with carrier ``'base'`` for all buses of a specific carrier, e.g.
+        ``'H2_ind_load'``. Connections (via links) of the selected buses to
+        buses of the carrier specified in ``'skip'`` are ignored in the
+        clustering and aggregated based on the newly clustered buses in a
+        following step.
+
```
ulfmueller, line 320: Sorry, from this explanation I do not get the 'skip' functionality. Can you explain it to me? And maybe it is worth it to reformulate the given definition - at least if you wanna reach the DAU ;)

ulfmueller (commented):
I am checking the results what I am getting with this branch (and its appl settings) for the SH model. 
What I am not getting right now or where I still need explanation:

- If there is 5 H2_grid buses - how can H2_ind_load be 9 ? Are there different types of H2_ind_loads which can't be aggregated. Otherwise I would expect to have, in terms of number of buses: H2_ind_loads <= H2_grid 
- In the case of central_heat and rural_heat, they get clustered to each 4 identities but being based to a 17 buses CH4 grid. Why is that? Are there CH4 buses with no central_heat or rural_heat clusters?

etrago/appl.py
fwitte: I'll try to rephrase that, thank you for the comment.

The idea is, that the `bus0` of all links (e.g. heat pump, gas boiler, resistive heater, ...) connected to the sector coupled buses (e.g. the central heat buses) are collected. Then, for each sector coupled bus, I check which other buses have the same combination of `bus0` on the connected links. Buses with the same combination are then aggregated. For this, the storage buses have to be ignored/skipped in the first step, since every heat bus has an own storage bus, which does mean that the combination of `bus0`s including the storage bus is unique for every heat bus.

Therefore, after the aggregation of the heat buses, the heat storages are aggregated according to the newly created busmap.

fwitte:
@ulfmueller, can we meet tomorrow or on Wednesday to discuss your questions?

Heads up:

>     * If there is 5 H2_grid buses - how can H2_ind_load be 9 ? Are there different types of H2_ind_loads which can't be aggregated. Otherwise I would expect to have, in terms of number of buses: H2_ind_loads <= H2_grid

That is correct. However, I do get 6 H2_ind_load buses in SH before clustering and 3 after clustering (with 4 H2_grid buses in SH).

>     * In the case of central_heat and rural_heat, they get clustered to each 4 identities but being based to a 17 buses CH4 grid. Why is that? Are there CH4 buses with no central_heat or rural_heat clusters?

After clustering I do get 4 CH4 buses in SH and one in the middle of Germany (is only implemented for test mode), all other CH4 buses are neighbouring countries. Maybe this is the issue?

The gas clustering only affects the network in Germany. The neighbouring countries are not clustered.

Thank you for reviewing! :)

etrago/appl.py
ulfmueller: I think I roughly understand it now. It sounds like, that the skipping is really s.th. which should be done always with respect to the strategy (e.g. 'central_heat': {'base': ['CH4']})? Do you really want to give this skip option to the user or is it s.th. rather to adjust in the function itself because it is rather a theoretic thing to do for advanced users? At least I do not get when you would not want to use the skipping function...

ulfmueller (approved):
I did some mistakes in my analysis. Thanks for your explanations - I get the same 'numbers' and understand your reasoning.
I did not review the resulting system in depth, but as far as I checked it, I am ok with it. Thanks for you effort! 
(The skipping discussion is probably not that important and/or can be dealt with later)

etrago/appl.py
fwitte: Good point, I will remove the option and just handle it directly.

fwitte:
There is one last issue: While the busmap is generated correctly and the link bus0 and bus1 information are updated accordingly, aggregation of the links does not seem to take place. I will merge this PR after fixing it here.

fwitte:
Thanks @CarlosEpia, found the mistake, it was some kind of residual from our original implementation of the gas clustering. We simply did not aggregate non gas carriers.

Therefore, in this PR

- All link carriers are aggregated.
- All CH4 pipelines are considered bidirectional. That means, after clustering, if a pipeline exists with the bus combination `bus0=1111`, `bus1=2222` next to a pipeline with reversed order bus combination (`bus0=2222`, `bus1=1111`), they are aggregated to a single pipeline.

We need to keep this in mind, if we want to move aggregation to an independent step after creating the busmaps for all sectors.

PR #376, opened 2022-02-18
closes #348

### LINKED ISSUE #348: change the extendable design
So far, all the components you want to optimize investment-wise have to be called in the args, [here](https://github.com/openego/eTraGo/blob/dev/etrago/appl.py#L76) according to the possible parameters which will be processed in this [extendable function](https://github.com/openego/eTraGo/blob/dev/etrago/tools/extendable.py#L45)

The future data model will have, in a base-setting, already 'turned on' p_nom_extendable/s_nom_extendable (=True). Therefore the function shall be rewritten such that once a component which is optimized by default is called in the args, will take the True from the data model and once it is not mentioned in the args it has to be set to false. In addition, alternative optimizing strategies can be defined...

ulfmueller:


ClaraBuettner:
Looks good! Did you already tested it or should I do that?

ulfmueller:
I tested it and it worked for me. If you like you can give it a second look just to ensure I did not miss anything because in the end I have pretty happy that it could be adapted so easily... But yeah, you can also be time-efficient and just approve it without testing...

ClaraBuettner (approved):
I also tested this branch and it was working as expected

PR #385, opened 2022-02-24
Add option to use stored ch4-busmap for gas_clustering

KathiEsterl:
I added the functionality of using a given ch4-busmap for the gas_clustering to be able to reproduce the exact same network in different calculation runs. 

@AmeliaNadal , what do you think of this approach? Are you okay with this or do you have a better idea to tackle this?


AmeliaNadal (approved):
I think that this approach is really good, so in my opinion you can merge it like this.

ClaraBuettner (approved):

PR #386, opened 2022-02-24
Features/#371 update capital cost

### LINKED ISSUE #371: Update capital costs and annualization methods
Extension costs for all components are now part of eGon-data. In the current version of eTraGo, costs from eGon-data are over written in `set_line_costs` and  `set_trafo_costs`:

https://github.com/openego/eTraGo/blob/40f3d4b42a61f2832d1cae1455faa0a76bcb9b50/etrago/tools/utilities.py#L772-L856

I think I would like to keep the option of adjusting the costs assumptions in eTraGo. But we definitely have to avoid overwriting values from eGon-data. And we might also take into account capital costs from other sectors. 

In addition, the annualization function needs to be updated:
https://github.com/openego/eTraGo/blob/40f3d4b42a61f2832d1cae1455faa0a76bcb9b50/etrago/tools/utilities.py#L1093-L1131
This function needs some adjustments: 
- It only considers lines, links and storage_units by now, stores have to be added. 
- we have to decide if we have to set assumptions on life time (and interest rate) carrier-wise and integrate the new values
- costs for DC transformers are already added in eGon data, so this has to be removed here
- capital_costs for storage_units are not annualized in the database anymore

ClaraBuettner:
Closes #371 
This branch avoids overwriting of capital costs from the data model. 
The capital_costs are already annualized in eGon-data. `convert_capital_costs`function  is now used to calculate intra-year investment costs according to the number of selected snapshots. 
I tested this branch with a new data model for Schleswig-Holstein.

### DIFF
--- a/etrago/tools/extendable.py
+++ b/etrago/tools/extendable.py
@@ -103,8 +103,6 @@ def extendable(self, line_max):
             else:
                 network.links.p_nom_max = float("inf")
 
-        network = self.set_line_costs()
-        network = self.set_trafo_costs()
 
     if 'german_network' in self.args['extendable']:
         buses = network.buses[~network.buses.index.isin(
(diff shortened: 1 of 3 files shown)


ulfmueller:
Closes  #387 :laughing:

ulfmueller (approved):

PR #390, opened 2022-03-14
Drop time depended values in `drop_sectors`

ClaraBuettner:
The function `drop_sectors` didn't drop time-depended values of the dropped components.
In this branch the function is updated to also drop the time-depended values by using the pypsa-function `mremove` instead of removing entries from the DataFrame manually. 

@ulfmueller I was able to run the clustering with these changes.

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -1623,49 +1623,28 @@ def drop_sectors(self, drop_carriers):
 
     """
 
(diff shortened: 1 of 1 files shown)


ulfmueller (approved):
I didnt totally understand that mremove function from pypsa but the result looks great and I would trust that pypsa code then ;)
Great, @ClaraBuettner - Thanks for fixing this!

PR #401, opened 2022-03-24
Set number of gas clusters to specified number

### LINKED ISSUE #400: The number of CH4 buses after clustering is not equal to the specified number
The reason for this is, that the German CH4 buses are clustered independently of the neighboring country buses (which are not clustered at all). Currently, the number specified by the user indidcates the amount of German CH4 buses after clustering.

Proposed change: Specified amount of CH4 buses = neighboring country buses + German CH4 buses after clustering. A minimal amount of CH4 buses should therefore be introduced as well (number of neighboring country buses + 1 bus for Germany).

fwitte:
Resolve #400

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -76,7 +76,7 @@ args = {
     'network_clustering_kmeans': {
         'active': True, # choose if clustering is activated
         'n_clusters': 30, # number of resulting nodes
-        'n_clusters_gas': 5, # number of resulting nodes in Germany
+        'n_clusters_gas': 30, # number of resulting nodes in Germany
         'kmeans_busmap': False, # False or path/to/busmap.csv
         'kmeans_gas_busmap': False, # False or path/to/ch4_busmap.csv
         'line_length_factor': 1, #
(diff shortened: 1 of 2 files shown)


fwitte:
@ulfmueller: Should work as intended now, can you review? Thank you!

ulfmueller (commented):
Almost - but may be not quite ;) 
- must-have comment: resulting number of buses seems to be always n+1
- nice-to-have comment: could be nice to post the number of neighboring buses in the error message stating that the minimum number has to be higher than the number of foreign buses

fwitte:
[bcbc5a6](https://github.com/openego/eTraGo/pull/401/commits/bcbc5a6870322445f94cf686e140877847be9b82) addresses both changes

>     * must-have comment: resulting number of buses seems to be always n+1

Thank you for noticing, I already had that implemented and just forgot to stage the changes... :D

>     * nice-to-have comment: could be nice to post the number of neighboring buses in the error message stating that the minimum number has to be higher than the number of foreign buses

Good point

ulfmueller (approved):

Issue #405, opened 2022-03-29
Avoid numerical issues in LOPF


ClaraBuettner:
I also had some warnings regarding numerical issues, even if an optimal solution was found. I had a look into the lp file and found some possible reasons: 

- Some p_max_pu values of renewable generators are very small (e-28) and cause large bounds. I would suggest to find a maximum reasonable value and set all smaller values to 0. 
- Large capital costs of H2_to_power and power_to_H2 (e13) cause large objective coefficients. I don't know if that's is the expected magnitude. @fwitte Does that make sense to you?

fwitte:
I ran the clustering on DE, it seems the capital_cost information gets distorted in the AC clustering. I made exports of the network before and after gas clustering:

[links_not_clustered.csv](https://github.com/openego/eTraGo/files/8377995/links_not_clustered.csv)

In gas clustering, with the strategy `"mean"` (I updated it locally only) for all link capital cost, I do get the follwing results, which seem somewhat consistent with the input values from the network *after* AC `k-means clustering` *without* `ehv clustering`.

[links_clustered.csv](https://github.com/openego/eTraGo/files/8377999/links_clustered.csv)

ulfmueller:
@fwitte I have troubles to understand some parts of your findings. When you say you used the stragegy 'mean' locally, does that mean you added a line for the links' capital costs [here](https://github.com/openego/eTraGo/blob/dev/etrago/cluster/gasclustering.py#L530)?

fwitte:
Yes, that is what I added, because I was not sure what the default strategy would be.

Independent of that, the capital cost information are already skewed after k-means clustering of AC. Unfortunately, I was not yet able to find, what is happening there.

fwitte:
The function

https://github.com/openego/eTraGo/blob/23085ae4c5366a0619825fb0a1aba3cbfdcc4538/etrago/tools/utilities.py#L1113-L1115

called by

https://github.com/openego/eTraGo/blob/23085ae4c5366a0619825fb0a1aba3cbfdcc4538/etrago/appl.py#L404

with these settings

https://github.com/openego/eTraGo/blob/23085ae4c5366a0619825fb0a1aba3cbfdcc4538/etrago/appl.py#L59-L60

does the trouble.

fwitte:
Update to my previous comment:

[...]

PR #409, opened 2022-04-07
Fix capital cost update

ClaraBuettner:
Capital costs should be smaller when not all snapshots are considered. I did a mistake in the previous implementation, sorry!

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -1108,23 +1108,23 @@ def convert_capital_costs(self):
 
     network.lines.loc[
         network.lines.s_nom_extendable == True, "capital_cost"
-        ] *= 8760 / n_snapshots
+        ] *= n_snapshots / 8760
 
     network.links.loc[
         network.links.p_nom_extendable == True, "capital_cost"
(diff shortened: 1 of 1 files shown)


ulfmueller:
Hi Clara, check out #405 ! I think there was a commit by @CarlosEpia addressing s.th. similar. But I am not sure - just to check whether there is some overlap...

CarlosEpia:
There were two problems with the capital:
1. In the function "convert_capital_costs", which Clara solved with this PR
2. in the function "group_links" which is solved in PR #410

So I think that there is no overlapping.

ClaraBuettner:
> There were two problems with the capital:
> 
>     1. In the function "convert_capital_costs", which Clara solved with this PR
> 
>     2. in the function "group_links" which is solved in PR [Fix/#405 avoid numerical issues in lopf #410](https://github.com/openego/eTraGo/pull/410)
> 
> 
> So I think that there is no overlapping.

Thank you for you comment. If it's fine for you, I will merge this branch into dev.

CarlosEpia (approved):
Works for me. It can be merged to dev.

PR #410, opened 2022-04-07
Fix/#405 avoid numerical issues in lopf

CarlosEpia:
Solves #405

### DIFF
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -946,6 +946,7 @@ def kmean_clustering(etrago):
             generator_strategies=strategies_generators(),
             aggregate_one_ports=aggregate_one_ports,
             line_length_factor=kmean_settings["line_length_factor"],
+            scale_link_capital_costs=False
         )
         network = clustering.network
 
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -946,7 +946,6 @@ def kmean_clustering(etrago):
             generator_strategies=strategies_generators(),
             aggregate_one_ports=aggregate_one_ports,
             line_length_factor=kmean_settings["line_length_factor"],
-            scale_link_capital_costs=False
         )
         network = clustering.network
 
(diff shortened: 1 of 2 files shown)


ClaraBuettner:
I run eTraGo successfully with the network for Germany. But I'm not completely sure what this PR is about. 
However, I think it is not solving all numerical issues, so I would make sure that the issue is not closed by this PR.

CarlosEpia:
eTraGo is running successfully, but there is a problem with the links' capital costs. When some links are aggregated, the resulting capital cost decreases. This PR solves that small problem.

etrago/cluster/networkclustering.py
```diff
@@ -985,8 +985,9 @@ def normed(x):
         generator_strategies=strategies_generators(),
         aggregate_one_ports=aggregate_one_ports,
         line_length_factor=kmean_settings["line_length_factor"],
+        scale_link_capital_costs=False
```
ClaraBuettner, line 988: Sorry for asking again, but do you think this change makes sense for all link carrier? Looking at PyPSA, this turns off the scaling of capital costs to the clustered link length. Is that correct? 
Setting it to False for links without a 'real length' like e.g. power_to_H2 makes sense for me. But for links with a real length e.G. DC-lines, the new length should be considered in the capital_costs. What do you think?

ClaraBuettner, line 988: Sorry for asking again, but do you think this change makes sense for all link carrier? Looking at PyPSA, this turns off the scaling of capital costs to the clustered link length. Is that correct? 
Setting it to False for links without a 'real length' like e.g. power_to_H2 makes sense for me. But for links with a real length e.G. DC-lines, the new length should be considered in the capital_costs. What do you think?

CarlosEpia:
For links with carrier H2_to_power and power_to_H2, the column "length" does not contribute to our calculations but creates numerical problems. Setting these values to 0 solves avoid this problem. One option is to delete these values in eGon-data, and the other is to do it in eTraGo (provisionally implemented in this PR). We agreed with @AmeliaNadal to discuss it during the project meeting.

Issue #412, opened 2022-05-03
Store the busmaps used for clustering


CarlosEpia:
In order to have the necessary data to develop the disaggregation methods, the busmaps used during the different clustering must be stored. A final busmap relating the original input and the final clustering must be created.

PR #417, opened 2022-05-09
Features/#412 store busmaps used for clustering

CarlosEpia:
Solve #412  and #410


CarlosEpia:
This PR has been modified in the way that it does not depend on csv files.

etrago/appl.py
```diff
@@ -1,496 +0,0 @@
...
-    print(datetime.datetime.now())
-    etrago.session.close()
-    # plots
-    # make a line loading plot
-    # plot_line_loading(network)
-    # plot stacked sum of nominal power for each generator type and timestep
-    # plot_stacked_gen(network, resolution="MW")
-    # plot to show extendable storages
-    # storage_distribution(network)
-    # extension_overlay_network(network)
```
ClaraBuettner, line 496: This looks a bit strange. Did you accidentally deleted the appl.py?

ClaraBuettner:
One general comment: It looks like there are many changes caused by using `black`. In eTraGo, we used to use another tool/style called `pep8`. We could also agree on using `black`, but I would like to discus that first. If we want to change it, it would make sense to apply black in separate commits, otherwise it is a bit challenging to understand what was changed in the commit :)

CarlosEpia:
I considered it easier and more comprehensible to create a new branch and PR #431 without black and the first implementation which used the csv files.

PR #418, opened 2022-05-10
Features/add line max abs

ulfmueller:
Upper bounds for grid expansion can be set by specifying so in the args...

### DIFF
--- a/etrago/tools/extendable.py
+++ b/etrago/tools/extendable.py
@@ -86,7 +86,8 @@ def extendable(self, grid_max_D= None,
         network.generators.p_nom_extendable = False
         
     if not grid_max_abs_D==None:
-        buses = network.buses[network.buses.country=='DE']
+        buses = network.buses[(network.buses.country=='DE') 
+                              & (network.buses.carrier == 'AC')]
         
         line_max_abs(network=network, buses=buses,
                      line_max_abs=grid_max_abs_D)
@@ -98,7 +99,9 @@ def extendable(self, grid_max_D= None,
                   'p_nom_max'] = grid_max_abs_D['dc']
         
     if not grid_max_abs_foreign==None:
-        foreign_buses = network.buses[network.buses.country!='DE']
+        foreign_buses = network.buses[(network.buses.country!='DE')
+                                      & (network.buses.carrier == 'AC')]
+
         line_max_abs(network=network, buses=foreign_buses,
                      line_max_abs=grid_max_abs_foreign)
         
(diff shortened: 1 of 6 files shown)


etrago/tools/extendable.py
```diff
@@ -66,56 +66,93 @@ def extendable(self, line_max):
...
+        network.lines.loc[(network.lines.bus0.isin(buses.index)) &
+                    (network.lines.bus1.isin(buses.index)),
+                    's_nom_max'] = grid_max_D * network.lines.s_nom  
+        
+        network.transformers.loc[network.transformers.bus0.isin(
+            buses.index),'s_nom_max'] = grid_max_D * network.transformers.s_nom
+        
+        network.links.loc[(network.links.bus0.isin(buses.index)) &
+        (network.links.bus1.isin(buses.index)),
+          'p_nom_max'] = grid_max_D * network.links.p_nom
```
ClaraBuettner, line 116: This sets upper limits for all links in Germany, right? To exclude the sector-coupling links, the` buses` in line 105 could get an additional filter `carrier == 'AC'`

etrago/tools/plot.py
```diff
@@ -1357,15 +1498,15 @@ def plot_grid(self,
     if disaggregated:
         network = self.disaggregated_network.copy()
     else:
-        network = self.network.copy()
+        network = self.copy()
```
ClaraBuettner, line 1501: I always run into errors with this change. Since `self` is an Etrago-object, the PyPSA-network is stored in `self.network`. Could you explain why you changed it?

etrago/tools/extendable.py
```diff
@@ -323,6 +298,85 @@ def extendable(self, line_max):
     return network
 
 
+def line_max_abs(network, buses, line_max_abs= {'380': 1000, '220': 700, '110':500, 'dc':2000}):
```
ClaraBuettner, line 301: Just out of curiosity, would you recommend to use these values? They seem a bit small, a default 380kV line already has a capacity of 1790MW...

etrago/tools/extendable.py
ulfmueller: Good point! Forgot about the other sectors :flushed: I will adjust that!

etrago/tools/extendable.py
ulfmueller: Good point! No, those values are not recommendable. I actually have a function for that upper bound definition, which I will integrate here as well. Thanks for bringing this up!

etrago/tools/extendable.py
ulfmueller: solved by d1bcb57b1b2bd48f95adf08ce183acae606315ba

etrago/tools/plot.py
ulfmueller: normalized again in 038a0df6563d21066d764ec7a663ee2bd4dad362

ClaraBuettner (approved):

Issue #415, opened 2022-05-05
Add e-moility carriers to clustering functions


ClaraBuettner:
The most recent egon-data backup includes e-mobility data. This currently leads to problems in the clustering because the carriers are not included in the functions yet. 
@CarlosEpia: Could you maybe have a look at this?

CarlosEpia:
> The most recent egon-data backup includes e-mobility data. This currently leads to problems in the clustering because the carriers are not included in the functions yet. @CarlosEpia: Could you maybe have a look at this?

Sure. I will manage this today.

PR #423, opened 2022-05-11
Features/#415 add emobility carriers to clustering functions

CarlosEpia:
Fixes PR #415

### DIFF
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -89,10 +89,10 @@ def adjust_no_electric_network(network, busmap, cluster_met):
         & (network2.buses["carrier"] != "central_heat")
         & (network2.buses["carrier"] != "central_heat_store")
     ]
-
     map_carrier = {
         "H2_saltcavern": "power_to_H2",
         "dsm": "dsm",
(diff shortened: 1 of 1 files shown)


etrago/cluster/networkclustering.py
```diff
@@ -261,6 +261,7 @@ def strategies_one_ports():
             "e_nom": np.sum,
             "e_nom_min": np.sum,
             "e_nom_max": np.sum,
+            "e_initial": np.mean,
```
ClaraBuettner, line 264: Energies are aggregated, so I think you should use `np.sum` instead of `np.mean`. What do you think?

ClaraBuettner:
I tested this branch and run eTraGo successfully. I think it can be merged as soon as the comment above is resolved.

ClaraBuettner (approved):

etrago/cluster/networkclustering.py
CarlosEpia: Agree. I assumed wrongly that "e_initial" was in p.u. as well as "e_nom_min" and "e_nom_max". It was corrected in b2beae8b3b28f1e2c11918058da5b524bfc779fd.

PR #431, opened 2022-05-20
Features/#412 store busmaps used for clustering v2

CarlosEpia:
Solve #412. Replace PR #417.

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -1734,3 +1734,25 @@ def drop_sectors(self, drop_carriers):
             two_port.df[
                 ~two_port.df.bus1.isin(self.network.buses.index)].index,
             )
+
+
+def update_busmap(self, new_busmap):
+    """
+    Update busmap after any clustering process
+    Parameters
+    ----------
+    new_busmap : dictionary
+        busmap used to clusted the network.
+    Returns
+    -------
+    None.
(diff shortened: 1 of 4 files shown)


etrago/tools/utilities.py
```diff
@@ -1734,3 +1734,25 @@ def drop_sectors(self, drop_carriers):
...
+    """
+    
+    if not self.busmap:
+        self.busmap = new_busmap
+    else:
+        for key in self.busmap.keys():
+            try:
+                self.busmap[key] = new_busmap[self.busmap[key]]
+            except:
+                print(f"bus {key} can not be mapped")
```
ClaraBuettner, line 1758: Do you think it makes sense to not throw error messages when a bus couldn't be mapped to the busmap? 

Besides, just as a nice-to-have or for future implementations: I think it is possible to implement this without a loop which would be faster.

ClaraBuettner:
I tested your branch and it worked as expected. When the comment above is solved, you can merge this.

etrago/tools/utilities.py
CarlosEpia: I deleted the error messages. Now, in case a bus can not be mapped, the full process will stop. The update of the busmap was improved in d348ddee18f40d9d61b2d46cca54703de6b41fb5

PR #433, opened 2022-05-24
Change import of time series data

ClaraBuettner:
This change should decrease the RAM usage during the import of timeseries.

### DIFF
--- a/etrago/tools/io.py
+++ b/etrago/tools/io.py
@@ -254,41 +254,51 @@ class NetworkScenario(ScenarioBase):
                     egon_etrago_transformer_timeseries
                     )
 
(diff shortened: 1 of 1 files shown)


CarlosEpia:
I tested this branch on my machine and the max use of memory RAM while building the network was about 28% (32Gb available).  It is a good improvement in comparison with the 68% used when using the current code available in dev. Additionally, the required time decreased significantly. Now it takes about a minute. The tables inside etrago.network look as expected.
The test was done using data for the entire Germany. This PR can be merged into dev.

CarlosEpia (approved):
It was tested successfully. Can be merged into dev.

Issue #432, opened 2022-05-23
AC clustering should have the options to include or not the cross-border buses


CarlosEpia:
The gas clustering has an option to exclude the cross-border buses from the clustering. The same characteristic should be implemented for the AC clustering (eHV and kmean).

PR #436, opened 2022-06-02
Optional clustering of cross-border AC buses

CarlosEpia:
Solve #432


CarlosEpia:
@ClaraBuettner this was implemented to have the same option that the gas sector has. But there is a difference that I suppose is worth discussing. There are in total 88 foreign AC buses, which means 88 extra buses to perform the optimization. It is a significant number if we keep in mind that many times we use a total of 300 buses in previous calculations.

ClaraBuettner:
> @ClaraBuettner this was implemented to have the same option that the gas sector has. But there is a difference that I suppose is worth discussing. There are in total 88 foreign AC buses, which means 88 extra buses to perform the optimization. It is a significant number if we keep in mind that many times we use a total of 300 buses in previous calculations.

I agree, we should discuss this. 88 foreign AC-buses seems a lot, there are much less foreign countries. So there is more than one bus per foreign country, right? Do you have a map of the 88 foreign buses you could post here?

CarlosEpia:
> > @ClaraBuettner this was implemented to have the same option that the gas sector has. But there is a difference that I suppose is worth discussing. There are in total 88 foreign AC buses, which means 88 extra buses to perform the optimization. It is a significant number if we keep in mind that many times we use a total of 300 buses in previous calculations.
> 
> I agree, we should discuss this. 88 foreign AC-buses seems a lot, there are much less foreign countries. So there is more than one bus per foreign country, right? Do you have a map of the 88 foreign buses you could post here?

Here we can see 84 of the 88 buses that have a country assigned different than "DE":
image
The other 4 buses have no country assigned, which is an issue that also requires a decision.
image

CarlosEpia:
About the 4 AC buses with no country data:
1. They are allocated on the North Sea:
2. Bus IDs: ['14523', '14398', '14233', '13590']
3. They only have connections to lines. No link, generator, load, or storage_unit is connected to them.
image

etrago/appl.py
```diff
@@ -320,8 +322,8 @@ def run_etrago(args, json_path):
                  by carrier, set upper/lower limit in p.u.
 
      network_clustering_kmeans : dict
-          {'active': True, 'n_clusters': 30, 'n_clusters_gas': 30,
-           'kmeans_busmap': False, 'line_length_factor': 1.25,
+          {'active': True, 'n_clusters': 30, 'cluster_foreign_AC': False,
+           'n_clusters_gas': 30, 'kmeans_busmap': False, 'line_length_factor': 1.25,
```
ClaraBuettner, line 326: The new argument "cluster_foreign_AC" is part of the default arguments but not described in the text below. Maybe it could be combined with the gas_clustering documentation

etrago/cluster/networkclustering.py
```diff
@@ -910,6 +922,8 @@ def normed(x):
         .set_index("T" + trafo_index),
         "Line",
     )
+    network.lines.lifetime = 40
```
ClaraBuettner, line 925: Is this line still needed?

etrago/appl.py
CarlosEpia: Description included in adf6dff52902271495153c290ed2c01ba9ed0ce6

etrago/cluster/networkclustering.py
CarlosEpia: Fixed in adf6dff52902271495153c290ed2c01ba9ed0ce6

PR #441, opened 2022-06-14
Updated deprecated from_csv to pandas.read_csv()

pieterhexen:
https://github.com/openego/eTraGo/blob/ed40d087da2d1164d2474167fb59c877469b063e/etrago/cluster/gasclustering.py#L150 
was added to make sure that every bus has a related weight (value = 1) even if it is not available in the specified .csv file.


etrago/cluster/gasclustering.py
```diff
@@ -146,8 +139,14 @@ def weighting_for_scenario(ch4_buses, save=None):
             save="network_ch4_" + kmean_gas_settings["bus_weight_tocsv"],
         )
     elif kmean_gas_settings["bus_weight_fromcsv"] is not None:
-        weight_ch4 = pd.Series.from_csv(kmean_gas_settings["bus_weight_fromcsv"])
-        weight_ch4.index = weight_ch4.index.astype(str)
+        # create DataFrame with uniform weightings for all ch4_buses
+        weight_ch4 = pd.DataFrame([1] * len(buses_ch4), index=buses_ch4.index)
+        loaded_weights = pd.read_csv(
+            kmean_gas_settings["bus_weight_fromcsv"], index_col=0
```
AmeliaNadal, line 145: I think it is a problem that the parameter bus_weight_fromcsv is the same for the kmean_clustering and the gas clustering. In my opinion, it would be easier to have to different parameters in the arguments.

pieterhexen: > I think it is a problem that the parameter bus_weight_fromcsv is the same for the kmean_clustering and the gas clustering. In my opinion, it would be easier to have to different parameters in the arguments

I agree and updated everything accordingly in 23ad144c2aa11bffc11ab85e2742236219b657b8

AmeliaNadal (approved):
This branch is working fine and can be merged into dev.

PR #456, opened 2022-07-01
Feature/update n clusters gas

pieterhexen:
Updated and tested everything according to our discussion: n_clusters_gas now refers to the total number of resulting buses regardless of wether foreign nodes are considered or not.


pieterhexen:
Also applied black to appl.py

pieterhexen:
@CarlosEpia Found the problem! I fixed/updated it already some time ago. Can you review/approve so I can simply merge?

CarlosEpia (approved):
The changes look good and it works as expected.

PR #458, opened 2022-07-01
Refer to correct db specification

fwitte:
Fix https://github.com/openego/eTraGo/issues/457

The `etrago.args` attribute is overwritten in line 124 of etrago/tools/network.py but the database selection refers to the default args.


CarlosEpia (approved):
Thanks for noticing Francesco. It makes sense to merge this change.

PR #469, opened 2022-07-15
added partial args loading and applied black

pieterhexen:
Implemented your solution. Tests look good.


pieterhexen:
@fwitte Relevant changes (not black-related) are in utilities.py line 1284-1299

fwitte:
Cool, I'll have a look later ;)

fwitte (approved):
Everything worked as expected for me. I think it is a good feature, what do you think @ClaraBuettner?

PR #476, opened 2022-07-28
Fix bug with too many clusters within k-medoids Dijkstra Clustering

KathiEsterl:
Set minimal weighting for spatial clustering and simplify identification of medoid per cluster in k-medoids Dijkstra Clustering

### DIFF
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -765,14 +765,18 @@ def busmap_from_psql(etrago):
     if not busmap:
         print("Busmap does not exist and will be created.\n")
 
-        cpu_cores = input("cpu_cores (default 4): ") or "4"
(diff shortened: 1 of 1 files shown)


CarlosEpia (approved):
Looks good to me. This branch can be merged.

KathiEsterl:
I added the query to set the cores for the multiprocessing. 

@CarlosEpia Maybe you want to have another look and give me a sign, then I will merge this branch. 

Thank you!

CarlosEpia (approved):
It looks good to me. Do you think that it could be useful to include in the message the maximum possible number of cores? Something like "cpu_cores (default 4, max {mp.cpu_count()}): "

KathiEsterl:
> It looks good to me. Do you think that it could be useful to include in the message the maximum possible number of cores? Something like "cpu_cores (default 4, max {mp.cpu_count()}): "

I added this option. Thank you!

PR #479, opened 2022-07-28
consider s_nom_min when calculating s_nom_max for transformers

### LINKED ISSUE #478: s_nom_min greater than s_nom_max in some transformers
When calculating the s_nom_max for transformers, there are some cases when s_nom_min is greater than s_nom_max. The function where s_nom_max is calculated was checked and it actually makes sense, but it ignores completely the value in s_nom_min. It leads to infeasibility when running optimizations.

CarlosEpia:
Fix #478.

### DIFF
--- a/etrago/tools/extendable.py
+++ b/etrago/tools/extendable.py
@@ -379,7 +379,10 @@ def line_max_abs(network, buses, line_max_abs= {'380':{'i':1020, 'wires':4, 'cir
 
 
 def transformer_max_abs(network, buses):
-
+    
+    # To determine the maximum extendable capacity of a transformer, the sum of
+    # the maximum capacities of the lines connected to it is calculated for each
+    # of its 2 sides. The smallest one is selected.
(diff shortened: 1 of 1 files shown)


CarlosEpia:
@ClaraBuettner these changes just guarantee that s_nom_min is <= s_nom_max. It is possible that some changes are needed to improve the calculation of s_nom_min (or not), but they should be developed in eGon-data.

ClaraBuettner:
I had a brief look at osmtgmod which sets the initial values for s_nom / s_nom_min. It seems like it selected the maximum sum of incoming / outgoing lines at the corresponding buses. So it seems to be reasonable that it is sometimes bigger than the s_nom_max calculated here. 
Could you maybe add a short comment before your additional line of code to explain what you do?

CarlosEpia:
I just added some comments to explain briefly the procedure to calculate the max capacity and the new line to avoid the problem described in #478

ClaraBuettner (approved):
Looks good, thank you!

Issue #473, opened 2022-07-21
Dispensable AC buses before clustering


CarlosEpia:
We use OSM data to create our data model. It includes some buses that are just in the middle of other two buses, but they don't have any load, generators or any other kind of function. They can be eliminated before performing any kind of clustering or optimization.

PR #483, opened 2022-08-15
Features/#473 dispensable ac buses before clustering

CarlosEpia:
Solve #473


pieterhexen:
Hey @CarlosEpia ,
After testing i found that the number of AC nodes goes down from more than 13k nodes to a little more than 9000 nodes when using `etrago.adjust_network()`. After checking the remaining nodes i saw that there are still 5003 nodes left that only have 'Line_AC' and 'StorageUnit_battery' attached to them. These should be removed aswell, right?

pieterhexen:
Everything looks to be working as intended!

Issue #466, opened 2022-07-14
eHV clustering is not working after the Dijkstra clustering was merged


CarlosEpia:
In order to integrate the new Dijkstra clustering, some incompatible changes with the existing eHV clustering were included. All the clustering methods must be able to coexist and run successfully when required.

CarlosEpia:
Solved in #472

PR #484, opened 2022-08-15
Fixes/#466 e hv clustering not working after merging dijkstra clustering

CarlosEpia:
Solves #466.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -119,7 +119,7 @@ args = {
             "rural_heat": {"base": ["CH4", "AC"], "strategy": "consecutive"},
         },
     },
-    "network_clustering_ehv": True,  # clustering of HV buses to EHV buses.
+    "network_clustering_ehv": False,  # clustering of HV buses to EHV buses.
     "disaggregation": "uniform",  # None, 'mini' or 'uniform'
     "snapshot_clustering": {
         "active": False,  # choose if clustering is activated
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -119,7 +119,7 @@ args = {
             "rural_heat": {"base": ["CH4", "AC"], "strategy": "consecutive"},
         },
     },
-    "network_clustering_ehv": False,  # clustering of HV buses to EHV buses.
+    "network_clustering_ehv": True,  # clustering of HV buses to EHV buses.
     "disaggregation": "uniform",  # None, 'mini' or 'uniform'
     "snapshot_clustering": {
         "active": False,  # choose if clustering is activated
--- a/etrago/cluster/networkclustering.py
+++ b/etrago/cluster/networkclustering.py
@@ -410,7 +410,6 @@ def cluster_on_extra_high_voltage(etrago, busmap, with_time=True):
     """
 
     network_c = Network()
-
     network, busmap = adjust_no_electric_network(etrago, busmap, cluster_met="ehv")
 
     pd.DataFrame(busmap.items(), columns=["bus0", "bus1"]).to_csv(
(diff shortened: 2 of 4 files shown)


etrago/cluster/networkclustering.py
```diff
@@ -781,23 +781,81 @@ def fetch():
...
+                    
+                    ----------------------- WARNING ---------------------------
+                    THE FOLLOWING BUSES WERE DELETED BECAUSE THEY WERE ISOLATED:
+                        {delete_buses.to_list()}.
+                    IT IS POTENTIALLY A SIGN OF A PROBLEM IN THE DATASET
+                    ----------------------- WARNING ---------------------------
+                    
+                    """)
+                    
+    network.buses = network.buses.drop(delete_buses)
```
ClaraBuettner, line 826: These lines are working, but in general I would recommend to use the build-in pypsa functions to drop components (see e.g. https://github.com/openego/eTraGo/blob/ad10466425386b2aa42a6f78c4f03d562188783d/etrago/tools/utilities.py#L1712-L1715). When you use them, e.g. time series are dropped automatically.

etrago/cluster/networkclustering.py
```diff
@@ -781,23 +781,81 @@ def fetch():
...
         busmap = busmap_from_psql(self)
 
         self.network, busmap = cluster_on_extra_high_voltage(
-            self.network, busmap, with_time=True
+            self, busmap, with_time=True
         )
 
         self.update_busmap(busmap)
-
+        buses_by_country(self)
```
ClaraBuettner, line 858: I think this should be `self.buses_by_country()` , right?

etrago/cluster/networkclustering.py
```diff
@@ -1409,9 +1471,13 @@ def kmedoids_dijkstra_clustering(etrago):
...
 
     if self.args["network_clustering"]["active"]:
+        
+        if self.args["network_clustering_ehv"]:    
+        
+            self.adapt_crossborder_buses()
 
         self.network.generators.control = "PV"
-
+
```
ClaraBuettner, line 1480: It looks like you should run `black` again...

ClaraBuettner:
Unfortunately I always run into errors when I try to use the kmedoids-dijkstra-clustering. I was only able to test eHV clustering, which was working as expected now. 
I would like to be able to test both methods before adding my review.

etrago/appl.py
```diff
@@ -119,7 +119,7 @@
             "rural_heat": {"base": ["CH4", "AC"], "strategy": "consecutive"},
         },
     },
-    "network_clustering_ehv": False,  # clustering of HV buses to EHV buses.
+    "network_clustering_ehv": True,  # clustering of HV buses to EHV buses.
```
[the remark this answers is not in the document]
CarlosEpia: network_clustering_ehv set to False again.

etrago/cluster/networkclustering.py
CarlosEpia: The function mremove is used now to delete elements.

etrago/cluster/networkclustering.py
CarlosEpia: buses_by_country() was aggregated to the eTraGo methods.

etrago/cluster/networkclustering.py
CarlosEpia: Black was used again.

CarlosEpia:
@ClaraBuettner I think this branch is ready for a second check. Since I merged dev into it, I expect that it works even with the dataset from the 16th of August.

PR #486, opened 2022-08-22
Replace NaN values with pypsa's default values during import

ClaraBuettner:
This branch replaces all NaN-values with the default values directly in the import of the network. This way, we can avoid setting these values manually in the application file.


CarlosEpia (approved):
Looks good to me. I merged dev into this branch locally and everything works as expected.

PR #493, opened 2022-09-02
updated simulataneous strategy

pieterhexen:
Hey @ClaraBuettner, we implemented a quick solution to improve the number of resulting heat buses after clustering. Can you test it a let me know if this is the result you were looking for? Thanks!

PR #494, opened 2022-09-06
Updated simultaneous strategy

pieterhexen:
@ClaraBuettner  In reference to #493 : 
I updated the simultaneous strategy now. Can you please check on if it shows the desired behaviour? If not I'm happy to adjust!


ClaraBuettner (approved):
Sorry for the late reply. The clustered network looks now as expected, thank you for the change!

PR #508, opened 2022-11-03
Gas dijkstra

pieterhexen:
As discussed this finishes the kmedoids-Dijkstra clustering for the CH4 grid.
The CH4 and H2_grid nodes are now moved to their correct location after clustering.


etrago/cluster/gas.py
```diff
@@ -51,21 +51,21 @@ def preprocessing(etrago):
...
     # select buses dependent on whether they should be clustered in (only DE or DE+foreign)
     if not kmean_settings["cluster_foreign_gas"]:
 
         network_ch4.buses = network_ch4.buses.loc[
-            ch4_filter
-            & (network_ch4.buses["country"].values == "DE")
+            ch4_filter & (network_ch4.buses["country"].values == "DE")
         ]
 
+        network_ch4.buses = network_ch4.buses.loc[~network_ch4.buses.index.isin(['48627', '48630'])]
```
CarlosEpia, line 67: Would it be a problem to cluster these buses to another country as the kmedoids-Dijkstra algorithm suggests?

AmeliaNadal:
The clustering is fonctionning well, when the problem with the two 'isolated' buses is solved, it could be merge.

PR #512, opened 2022-11-08
Add missing country

### LINKED ISSUE #511: Missing country in buses_by_country function
Russia is missing in the [countries dictionnary](https://github.com/openego/eTraGo/blob/dev/etrago/tools/utilities.py#L201:L216) of the buses_by_country function. This leads to the change of `country` of the CH4 bus in Russia from 'RU' to 'FI'.

AmeliaNadal:
Fixes #511


fwitte (approved):
Thank you!

Issue #495, opened 2022-10-11
Update pypsa version to 0.20.1


ClaraBuettner:
- [x] results are the same 
- [x] build_network_from_db()
- [x] adjust_network()
- [x] ehv_clustering()
- [x] spatial_clustering()
- [x] spatial_clustering_gas()
- [x] snapshot_clustering()
- [x] skip_snapshots()
- [x] lopf()
- [x] dispatch_disaggregation()
- [x] plots?

ClaraBuettner:
When the plots should work with cartopy, shapely needs to be installed without binaries:

`pip install shapely --no-binary shapely `

This seems to be necessary because otherwise shapely uses the system-wide version of GEO which could be different to the one in the virtualenv. 
It looks like there is no option to add this into the setup.py

Issue #426, opened 2022-05-16
Dependency missing in setup.py: cartopy


fwitte:
9bb242516462ef7bab5c8361d65b1f4c7e86817c brought cartopy in the dependencies, but it is not yet included in setup.py

ClaraBuettner:
The installation of `cartopy` could be a bit complicated and you also need to install system-wide packages. 
The tool is only used for background maps in plots. We could also make the usage of `cartopy` optional. That way it would be only used when it is installed, and eTraGo would also work without installing it. 
What do you think?

fwitte:
Making it optional is a good idea! I'll make some adjustments to #434 (remove from setup.py and adjusting the imports in plot.py).

PR #514, opened 2022-11-09
Update to pypsa version 0.20.1

ClaraBuettner:
Solves #495 
Solves #426


ClaraBuettner:
I tried to update the version of `matplotlib` as @pieterhexen suggested. But unfortunately I run into the same problem than before. 
So it looks like there is no other option than described here https://github.com/openego/eTraGo/issues/495#issuecomment-1307248517 . I will look for a good place in the documentation to add this solution there.

CarlosEpia (approved):
LGTM. I tested this branch and it works nicely. Can be merged into dev.

PR #530, opened 2022-11-18
Features/#507 plot powerto h2 expansion

Alonsoju95:


etrago/tools/plot.py
```diff
@@ -1565,8 +1565,9 @@ def plot_grid(self,
         link_colors = pd.Series(data=0, index=network.links.index)
     else:
         logger.warning("line_color {} undefined".format(line_colors))
-
+
```
ClaraBuettner, line 1568: There are some line with white space which are not needed. They have no effect, but we try to follow some code style rules which don't like these white spaces. 
You could use a tool called `black` to automatically delete them. Did you already used that in eGon-data? If not, I can show you how that works.

etrago/tools/plot.py
```diff
@@ -1578,18 +1579,27 @@ def plot_grid(self,
...
+        bus_sizes = network.storage_units.groupby(['bus', 'carrier']).p_nom_opt.sum()*bus_scaling
         bus_legend = 'Storage distribution'
-        bus_unit = 'TW'
+        bus_unit = 'TW'        
     elif bus_colors == 'gen_dist':
         bus_scaling = bus_sizes
         bus_sizes = bus_scaling * calc_dispatch_per_carrier(network, timesteps)
         bus_legend = 'Dispatch'
         bus_unit = 'TW'
+    elif bus_colors == 'PowerToH2':
```
ClaraBuettner, line 1592: Could you add a short description of this option to the doc string of the method? (line 1470)

etrago/tools/plot.py
Alonsoju95: I just pushed the changes

ClaraBuettner (approved):
Your PR looks good to me. You can merge this into dev-branch.

Issue #534, opened 2022-11-23
Update function plot_carrier


CarlosEpia:
The function plot_carrier was created to visualize quickly all kinds of buses and links. Some extra functionalities and format can improve the plot.

ClaraBuettner:
I this issue already solved?

CarlosEpia:
Yeah. I proceed to close it.

PR #535, opened 2022-11-23
Fixes/#534 update function plot carrier

CarlosEpia:
Solved #534

### DIFF
--- a/etrago/tools/plot.py
+++ b/etrago/tools/plot.py
@@ -1336,6 +1336,8 @@ def calc_network_expansion(network, method='abs', ext_min=0.1):
 
     """
     all_network = network.copy()
+    all_network.links = all_network.links[all_network.links.carrier == 'DC']
+    all_network.buses = all_network.buses[all_network.buses.carrier == 'AC']
     network.lines = network.lines[network.lines.s_nom_extendable &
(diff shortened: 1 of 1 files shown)


KathiEsterl:
I checked it (without cartopy as this is not working in my case) and everything seems to work fine. 

There are two minor things I would improve:
1) The title of the graph is "Links: ..." with the shown carriers. As there are also lines included (with the carrier 'AC', this might be confusing. I think I would just leave out the title or just say "Network with specific cariers". The shown carriers will be clear looking at the legend. 
2) I would name the AC-lines "Line AC" and not link, as those are lines in the network.

CarlosEpia:
> I checked it (without cartopy as this is not working in my case) and everything seems to work fine.
> 
> There are two minor things I would improve:
> 
> 1. The title of the graph is "Links: ..." with the shown carriers. As there are also lines included (with the carrier 'AC', this might be confusing. I think I would just leave out the title or just say "Network with specific cariers". The shown carriers will be clear looking at the legend.
> 2. I would name the AC-lines "Line AC" and not link, as those are lines in the network.

The title was deleted and the AC lines are named also like that in the legend.

KathiEsterl (approved):

Issue #533, opened 2022-11-22
Update sector_coupled_clustering in args


CarlosEpia:
There are several changes in the clustering of the sector-coupled buses:
            "H2_ind_load": **Does not exist anymore**
            "central_heat": {"base": ["CH4", "AC"], "strategy": **"simultaneous"**},
            "rural_heat": **now clustered based on AC buses**

Based on this, the arguments in appl.py and the necessary clustering functions should be updated.

PR #537, opened 2022-11-24
Features/#533 update sector coupled clustering in args

CarlosEpia:
Solves #533


pieterhexen (commented):
Tested and everything works fine!

PR #544, opened 2022-12-05
Import s_max_pu time series of lines in gas clustering

### LINKED ISSUE #542: Time series of lines get lost in clustering
After the clustering, there are not time-dependent values for transmission lines anymore. Before the clustering (after etrago.adjust_network) they are still there, so it looks like they get lost within the clustering.

This results in the problem that we can't see dynamic-line rating in the calculations.

ClaraBuettner:
Hi @AmeliaNadal and @pieterhexen 
This change seems to fix #542, could you have a look if it makes sense? 
I only added the s_max_pu timeseries of lines, we don't use time series of storage_units or transformers (yet), but if we would do so, they would be also deleted in that function. So it might be a good idea to automatically add al time series data of these components.

### DIFF
--- a/etrago/cluster/gas.py
+++ b/etrago/cluster/gas.py
@@ -367,6 +367,12 @@ def gas_postprocessing(etrago, busmap, medoid_idx):
 
     # Insert components not related to the gas clustering
     io.import_components_from_dataframe(network_gasgrid_c, etrago.network.lines, "Line")
+    io.import_series_from_dataframe(
+        network_gasgrid_c,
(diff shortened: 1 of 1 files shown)


ClaraBuettner:
Thank you @AmeliaNadal! The code looks good and works well. Do you think we can merge it?

AmeliaNadal:
Yes! I'd asked @pieterhexen to have look, but if you've tested it, I think this can be merged.

AmeliaNadal (approved):

PR #549, opened 2022-12-06
Fix p_nom_max agrregation with inf values

### LINKED ISSUE #548: NaN values in p_nom_max of links after clustering
The p_nom_max of some extendable links is NaN after clustering. 
It looks like it happens when only np.inf values are summed up. I will change this and do sth. similar to this: 
https://github.com/openego/eTraGo/blob/1bdf3082dd3479e8d632ea1851cfbc3e198d0e94/etrago/cluster/spatial.py#L99-L103

ClaraBuettner:
Fixes #548 
With this fix, the aggregation of np.inf values returns np.inf again instead of NaN. 

I'm sorry for the amount of PR I created in the last days... But could someone of you have a look at it @KathiEsterl, @AmeliaNadal or @pieterhexen ? Thank you in advance!

### DIFF
--- a/etrago/cluster/gas.py
+++ b/etrago/cluster/gas.py
@@ -19,7 +19,7 @@ if "READTHEDOCS" not in os.environ:
     from six import iteritems
 
     from etrago.cluster.spatial import (
-        agg_e_nom_max,
+        sum_with_inf,
         group_links,
         kmedoids_dijkstra_clustering,
     )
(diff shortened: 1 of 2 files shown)


pieterhexen:
Hey Clara, 
I did a little adjustment to gas.py to fit the new method name. The aggregation of np.inf values looks to be working as intended! 

Oddly the 'CH4' and all heat-related generators have a an empty p_nom_max and the h2_to_power links have integer values in p_nom_max. I will check it tomorrow and come back at you.

pieterhexen:
> Oddly the 'CH4' and all heat-related generators have a an empty p_nom_max and the h2_to_power links have integer values in p_nom_max. I will check it tomorrow and come back at you.

After talking to Amélia this turned out to be the right behaviour. So everything is fine!

PR #558, opened 2022-12-14
Reformat code and squelch a few `flake8` complaints

gnn:
Finally got around to creating a pull request with the cleanup I made while working on disaggregating the new sectors. Please have a look and make sure that I didn't make any breaking changes. One note: the `== False` and similar patterns is something that also cropped up in `egon-data` so I'm pointing it out here because it can easily be avoided. It's not really wrong. Just weird. :)


gnn:
Sorry, that was to early. I still have to squash some commits and do a force push.

gnn:
Squashed all temporary commits and fixed the conflicts by "rebasing" onto an updated "dev". The new code also uncovered that sometimes, using `== False` really is the wrong thing to do. ;) Have fun reviewing.

ClaraBuettner:
I had a brief look at the code changes which look fine for me. I tried to run this branch, but run into problems which are already fixed on dev but not part of this branch yet. 
When dev is merged into this branch, I will give it another try.

ClaraBuettner (approved):
I tried to run eTraGo on this branch and it worked as expected, so I think you can merge this branch

PR #559, opened 2022-12-14
Features/#525 lowflex scenario

### LINKED ISSUE #525: Test eGon2035_low_flex scenario
There will be a low_flex scenario of eGon2035 soon (?). 
When the data is there, we need to try out if eTraGo can handle this and if the results look as expected.

ClaraBuettner:
Fixes #525 
eTraGo needs some small changes to be compatible to the lowflex_scenarios. With this branch, it is possible to run calculations with eGon2035_lowflex


CarlosEpia (approved):
I just tested this branch with different scenarios and it works as expected.

Issue #541, opened 2022-12-02
Foreign buses which are part of a transformer not clustered properly 


CarlosEpia:
There are some foreign buses that are part of transformers that are currently clustered to buses inside Germany. It creates problems since it looks like a completely new AC line, but with the wrong characteristics.

PR #560, opened 2022-12-14
Solve problem while clustering foreign buses

CarlosEpia:
solves #541

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -435,7 +435,7 @@ def run_etrago(args, json_path):
 
     """
     etrago = Etrago(args, json_path)
-    
+
     # import network from database
     etrago.build_network_from_db()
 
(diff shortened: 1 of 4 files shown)


CarlosEpia:
This branch fixes the problem related to foreign buses clustered to German buses
image

CarlosEpia:
The problem with AC buses is also solved.
image

KathiEsterl:
I would take over the reviewing of this PR :)

pieterhexen:
> I would take over the reviewing of this PR :)

Great! I already applied the code in another branch and found that some 'DE' buses are isolated when only considering the German grid. What happend is that 'DE'-buses only connected through foreign nodes to the German grid are then isolated. The relevant bus-ids I found are '32461' and '33515'. The following image shows the initial state of bus '33515' at the DE/AT border:
grafik

I only applied the method form this PR after doing some other aggregation so it is possible, that the problem is tied to the the things I did. But maybe this serves as a hint for you on where to look a bit closer :)

PR #567, opened 2022-12-30
Fixes/#536 snapshots not converge in pf

CarlosEpia:
Solves 536


ClaraBuettner (approved):
Looks good to me and is working fine. It only looks like you should run black again on execute.py.

PR #572, opened 2023-01-15
Add a GitHub workflow

gnn:
Currently it only checks whether the package can be build without errors and whether files that had `black`, `flake8` and `isort` issued fixed did not regress. But it's really easy to add other checks, like e.g. unit tests.
The most important change here is that when you fix `black`, `flake8` and `isort` issues for a file, you can add it to the list of `cleaned` files in the `check` function in "noxfile.py" and the GitHub actions will make sure that it stays that way.


gnn:
I opened this PR to early, so the list of commits is too long. Closing so I can open another one with a shorter list of commits.

Issue #577, opened 2023-01-16
Exception Dijkstra clustering while unify_foreign_buses


CarlosEpia:
Sometimes the number of buses with loads/generators per foreign country is the same as the total number of buses. For this cases, the Dijkstra clustering can be omitted.

CarlosEpia:
Another function caused the problem. This issue was omitted.

PR #578, opened 2023-01-17
exception for Dijkstra clustering

CarlosEpia:
Solves #577


ClaraBuettner:
Hi @CarlosEpia, 
I tried to run your branch and add the OCGT in foreign countries (as in https://github.com/openego/eTraGo/commit/5c0d9226b3c541e05863a10e1863fc74168477b6). The clustering is now running without an error. 
But somehow the OCGT in foreign countries are lost after the clustering. Is that also happening in your calculations?

PR #586, opened 2023-01-27
Feature/vary h2feedin share

AmeliaNadal:


etrago/appl.py
```diff
@@ -75,6 +75,7 @@
     # Scenario variations:
     "scn_extension": None,  # None or array of extension scenarios
     "scn_decommissioning": None,  # None or decommissioning scenario
+    "H2_vol_share": 15,  # in % [20/15/10/5/2/0.1/0.01/0.001/0] allowed H2 volumetric share for feedin
```
ClaraBuettner, line 78: Could you please also add this change to args.json?

AmeliaNadal: Yes, I made it in https://github.com/openego/eTraGo/pull/586/commits/88218a91d9b376efdf861734c07619260808c953

AmeliaNadal:
We decided to unactivate H2_feedin for the paper calculations, so the changes of this PR are not needed anymore.

PR #590, opened 2023-02-02
Test package installation during CI

### LINKED ISSUE #589: Pinning `tsam == 1.1.0` leads to installation errors
In d02bc2fe the "tsam" package was [pinned to version `1.1.0`][0]. That version [depends on the "sklearn" package][1], which is deprecated, fails to install and thus also leads to [installation errors][2] with eTraGo. Can we unpin the "tsam" dependency, as to not pull in the broken "sklearn"?

[0]: https://github.com/openego/eTraGo/commit/d02bc2fe#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7
[1]: https://github.com/FZJ-IEK3-VSA/tsam/blob/d7ab4662ec08845b4c51356b74b1b7d177994510/requirements.txt#L1
[2]: https://github.com/openego/eTraGo/actions/runs/4071118757/jobs/7012615640#step:10:621

gnn:
This tests that the package can be installed successfully and thus ensures that whatever we do to fix issue #589 really fixes package installation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,3 +52,7 @@ jobs:
         run: "python -m nox -s build"
         env:
           PLATFORM: ${{ matrix.os.image }}
+      - name: "Install the package"
+        run: "python -m nox -s install"
+        env:
+          PLATFORM: ${{ matrix.os.image }}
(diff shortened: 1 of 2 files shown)


ClaraBuettner:
The version of tsam was unpinned in #628. I merged the updated `dev` into this branch to check if this fixes the package installation (at least it fixed it locally for one of our students).

ClaraBuettner (approved):

PR #591, opened 2023-02-02
Fix setting branch capacity factor with dynamic line rating

ClaraBuettner:
This should fix the problem with the (n-1)-security factor and dynamic line rating. 
@CarlosEpia Could you please check if it is now working as expected?


CarlosEpia (approved):
This commits solve the problem of DLR and capacity factor. Now the max use of a transmission line is exactly 1.05, which corresponds with the best case: branch_capacity_factor= 0.7 and DLR=1.5

Issue #593, opened 2023-02-03
Clustering of H2_grid buses relies on H2_feedin links


AmeliaNadal:
The H2_grid buses have the same localisation than the CH4 buses and their clustering is using the one of the CH4 grid. Therefore, the H2_feedin links (connecting each CH4 bus to its corresponding H2_grid bus) are used to establish the [correspondence](https://github.com/openego/eTraGo/blob/dev/etrago/cluster/gas.py#L218:L220).
For cases where H2_feedin is not present in the system or for the eGon100RE scenario, this should be changed.
An easy option would be to replace the "H2_feedin" links by "H2_to_CH4" links but there is maybe a more elegant way to do it.

See: https://github.com/openego/eTraGo/commit/1a8606971df05c968ac339102270691a44f6f3d5

pieterhexen:
WIP on [https://github.com/openego/eTraGo/tree/features/update_clustering](url)


TODO:

- [ ] Change H2_feedin dependency in gas clustering to sql-table based ch4_h2_mapping
- [ ] Add csv import and export for self.ch4_h2_mapping 
- [ ] Test

pieterhexen:
In oder to make the gas clustering work for the eGon100RE scenario the following lines need to be adapted to also account for H2_retrofit links:
https://github.com/openego/eTraGo/blob/79a4fda44aeece79c8306dec073f69e780f6c4cc/etrago/cluster/gas.py#L764-L773

Will be implemeted and tested when AC clustering is working for eGon100RE

pieterhexen:
Done in https://github.com/openego/eTraGo/pull/595

PR #595, opened 2023-02-09
Features/update clustering

pieterhexen:
Solves https://github.com/openego/eTraGo/issues/593


AmeliaNadal (approved):
Code looks good and the clustering functions how it should.
Can you please apply black to tools/ io.py, network.py and utilities.py before merging this branch into dev?

PR #598, opened 2023-02-10
Features/#597 prepare pf for disaggregation

### LINKED ISSUE #597: pf_post lopf and desaggregation
The desaggregation is currently not working after the pf, because some links have been transferred to generators before the pf. 
This needs to be also done for the disaggregated network. 
In addition, links which have been connected to different gas buses lead to generators with the same bus and carrier which is a problem in the desaggregation. They need to be aggregated in` import_gen_from_links()`

ClaraBuettner:
Fixes #597 
This branch adjusts the pf_post_lopf function to allow running the disaggregation afterwards. 

@CarlosEpia: I also applied black, that's why so many lines changes. If you want to see the real changes I did, you can have a look at the first two commits.

### DIFF
--- a/etrago/tools/execute.py
+++ b/etrago/tools/execute.py
@@ -22,7 +22,8 @@
 execute.py defines optimization and simulation methods for Etrago object.
 """
 import os
-if 'READTHEDOCS' not in os.environ:
+
+if "READTHEDOCS" not in os.environ:
     import time
     import logging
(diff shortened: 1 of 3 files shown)


CarlosEpia (approved):
I tested this branch with 50AC buses and 20Snapshots. It works as expected and can be merged into dev.

PR #601, opened 2023-02-13
Cluster p_nom_max for storage units with inf values

ClaraBuettner:


tillsilke (approved):
looks all good, no issues so far. All p_nom_max for storage units equals 'inf'.

Issue #354, opened 2022-01-07
Modeling of H2 feedin


fwitte:
# Issue

I will update this issue with different solutions. The final decision should be aimed for the end of January.

The H2 feedin into the CH4 grid should be restricted to a maximum share in the eGon2035 scenario.

# Possible Solutions

## ...

# Implications

## Data required

Depending on the method that will be implemented to model H2 feedin into the CH4 pipeline grid, additional data are required for the optimization.

### Solution 1

...

## Optimization model

Depending on the method that will be implemented to model H2 feedin into the CH4 pipeline grid, additional constraints are required for the optimization.

### Solutions 1

...

# Decisions

- What solution to implement?
- Where to put what part of the code?
-  ...?

ClaraBuettner:
I had another idea for the H2 feedin limitation, I hope it's fine to add this here: 

There are fixed gas loads per node (industry and rural heat buildings without a heat pump), so we already know in advance that at least this amount of CH4 has to be available at each node and time step. 
We could assume that the H2 feedin at each node and snapshot can't be more than X% (e.g. 15% vol.) than the fixed gas load. That way we ensure, that each fixed CH4 load can be supplied by maximum x% of H2.
It would be a conservative assumption, because there are actually also other gas demands which are part of the optimization (CHP, OCGT) and not considered in this limitation, but the resulting share would be always smaller than the selected limit. And they are most likely smaller that the rural heat loads. 

To model this, we can set a `p_max_pu` time series for the H2 feedin links which are calculated based on the CH4 loads. This could be easily done in eTraGo and might not effect the calculation time that much, we also use that for modelling other technologies.

PR #602, opened 2023-02-13
Limit H2 feedin to a share of CH4 load 

AmeliaNadal:
The solution proposed by Clara in https://github.com/openego/eTraGo/issues/354#issuecomment-1406535810 is implemented in this branch.

This is a very conservative approach to limit the H2 feedin into the CH4 grid, but we'd rather go for this option than for a modification of the methanisation efficiency that I also presented at the project meeting. Doing so allows to have the H2_feedin independant of the methanisation and to use H2_feedin without building methanisation at all for instance.
 
A lot of changes are due to black and the content changes are all contained in the commit https://github.com/openego/eTraGo/commit/d1077114080484703a8093d28aa0e821a9b6e007.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -28,6 +28,7 @@ the function etrago.
 import datetime
 import os
 import os.path
+
 import numpy as np
 
 __copyright__ = (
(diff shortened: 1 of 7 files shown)


etrago/appl.py
```diff
@@ -75,6 +75,7 @@
     # Scenario variations:
     "scn_extension": None,  # None or array of extension scenarios
     "scn_decommissioning": None,  # None or decommissioning scenario
+    "H2_vol_share": 15,  # in % [50/20/15/10/5/2/1/0] allowed H2 volumetric share for feedin
```
ClaraBuettner, line 78: Could you also add this parameter to the json file?

AmeliaNadal: Yes (https://github.com/openego/eTraGo/pull/602/commits/61d31a42699485681e33d620077ce15845e3cf36), sorry about forgetting this again!

AmeliaNadal:
> All checks failed in this PR. It could be fixed by merging the most recent `dev` into your branch. Could you try that out?

Yes, this looks better now :)

Issue #606, opened 2023-02-16
set_q_national_loads is returning just nan values


CarlosEpia:
The function set_q_national_loads() is returning just nan values.

PR #607, opened 2023-02-16
Features/#606 set q national loads is returning just nan values

CarlosEpia:
Solves #606

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -470,15 +470,14 @@ def set_q_foreign_loads(self, cos_phi=1):
 
     Parameters
     ----------
-    network : :class:`pypsa.Network
-        Overall container of PyPSA
+    etrago : :class:`etrago.Etrago
+        Transmission grid object
     cos_phi: float
         Choose ration of active and reactive power of foreign loads
 
     Returns
     -------
-    network : :class:`pypsa.Network
-        Overall container of PyPSA
+    None
 
     """
     network = self.network
@@ -510,6 +509,7 @@ def set_q_foreign_loads(self, cos_phi=1):
     network.loads_t["q_set"].columns = network.loads_t["q_set"].columns.astype(
         str
     )
+
 
 
 def connected_grid_lines(network, busids):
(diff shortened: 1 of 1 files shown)


etrago/tools/utilities.py
```diff
@@ -444,26 +444,25 @@ def set_q_national_loads(self, cos_phi=1):
...
+    # are affected
+    network.loads_t.q_set = pd.merge(network.loads_t.q_set, new_q_loads,
+                                     how="inner", right_index= True,
+                                     left_index=True, suffixes=("", "delete_"))
+    network.loads_t.q_set.drop(
+        [i for i in network.loads_t.q_set.columns if 'delete' in i],
+        axis=1, inplace=True)
 
 
 def set_q_foreign_loads(self, cos_phi=1):
```
ClaraBuettner, line 468: We talked about not setting a default value to these functions

etrago/tools/utilities.py
```diff
@@ -444,26 +444,25 @@ def set_q_national_loads(self, cos_phi=1):
...
-        network.loads.index[
-            network.loads.bus.astype(str).isin(national_buses.index)
-        ].astype(int),
-    ] = network.loads_t["p_set"].loc[
+    
+    # Calculate q national loads based on p and cos_phi
+    new_q_loads = network.loads_t["p_set"].loc[
         :,
         network.loads.index[
             network.loads.bus.astype(str).isin(national_buses.index)
```
ClaraBuettner, line 452: q is now calculated for all loads attached to 'AC' buses. That can have some side-effects: e.g. e-Mobility is a load attached to the AC nodes in the lowflex scenario but not in eGon2035. If we keep it like this, e-Mobility has a q demand in some scenarios, but not in all.

etrago/tools/utilities.py
CarlosEpia: Solved in f60f919195b09e366905618db5e0c4358144d932

etrago/tools/utilities.py
CarlosEpia: Fixed in b0fb292aa49241c015f6e2ba9b9384e36cfdd21b

ClaraBuettner:
Thank you for the changes, it looks good now I think. 
But all CI checks failed, could you have a look at that?

ClaraBuettner (approved):

Issue #564, opened 2022-12-21
Adapt export of busmaps while spatial clustering


KathiEsterl:
Several busmaps are exported during the spatial clustering. In specific settings, some files are overwritten. Additionally, the names are not clear.

CarlosEpia:
Hi @KathiEsterl. Should we close this issue, or is there something still ongoing?

KathiEsterl:
Oh right, I forgot to close this issue. 

@pieterhexen has a link to the branch to adapt the export of the gas busmaps accordingly, maybe we should have a new issue on that task? 

But I guess we can close this one.

CarlosEpia:
We can change the name of this issue to something more general or create a new one. Both are ok for me.

PR #609, opened 2023-02-20
changed busmap im- and export

pieterhexen:
Hey Kathi and Carlos,
Related to https://github.com/openego/eTraGo/issues/564

I updated the busmap im- and export for the gassector. Can you check if this goes along the changes you did in the electricity sector?

Thanks :)

### DIFF
--- a/etrago/cluster/spatial.py
+++ b/etrago/cluster/spatial.py
@@ -676,7 +676,7 @@ def dijkstras_algorithm(buses, connections, medoid_idx, cpu_cores):
     busmap = busmap_ind.map(mapping).astype(str)
     busmap.index = list(busmap.index.astype(str))
 
-    return busmap, busmap_ind
+    return busmap
 
 
 def kmedoids_dijkstra_clustering(etrago, buses, connections, weight, n_clusters):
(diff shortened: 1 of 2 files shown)


etrago/cluster/spatial.py
```diff
@@ -656,79 +689,60 @@ def kmedoids_dijkstra_clustering(etrago, buses, connections, weight, n_clusters)
...
+        busmap_ind.name = "medoid_idx"
 
-        distances = pd.DataFrame(
-            data=kmeans.transform(buses.loc[buses_i, ["x", "y"]].values),
-            index=buses_i,
-            dtype=object,
+        export = pd.concat([busmap, busmap_ind], axis=1)
+        export.index.name = "bus_id"
+        export.to_csv(
+            "kmedoids_gasgrid_busmap_"
```
KathiEsterl, line 743: It would be good if you could export the gas busmap in the postprocessing as this function to conduct the k-medoids Dijkstra Clustering is also used in the clustering of the AC-network, hence the busmap with this name would also be created during that process.

KathiEsterl:
Apart from the comment I made above, everything looks fine. Maybe you can change that. Afterwards I would give it another test run on my local machine and then approve it :) 
Thank you!

etrago/cluster/spatial.py
pieterhexen: Done! 
I also removed the return of `busmap_ind` in the dijsktra clustering because it is not used anywhere.

KathiEsterl (approved):

PR #610, opened 2023-02-21
make cartopy optional again

### LINKED ISSUE #596: Optional cartopy dependency disappeared
Cartopy was once an optional dependency, so the plots also worked without it. But that feature somehow disappeared on the current dev, now it is not working without cartopy anymore. That should be fixed. 
It might be helpful to have a look at the changes from commit https://github.com/openego/eTraGo/commit/45c2d00332aab52c0bdd8b17d63f0705a4637332

CarlosEpia:
Fixes #596

### DIFF
--- a/etrago/tools/plot.py
+++ b/etrago/tools/plot.py
@@ -21,15 +21,16 @@
 """
 Plot.py defines functions necessary to plot results of eTraGo.
 """
+from math import log10, sqrt
 import logging
 import os
+
 from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


CarlosEpia:
It looks like I did a lot of changes, but it is due to the use of isort and black. The real changes are in the first commit.

ClaraBuettner:
Did you already tried out the plots with and without cartopy?

CarlosEpia:
> 

Yes, I did. Is one of them not working for you?

PR #629, opened 2023-03-24
Disaggregate links and stores

gnn:
I'm finally done with picking the commits from `stashes/egon-disaggregation-tryouts`. Please check if it still works as well as on the other branch, then we can move it into `dev`.

### DIFF
--- a/etrago/cluster/disaggregation.py
+++ b/etrago/cluster/disaggregation.py
@@ -766,7 +766,6 @@ def run_disaggregation(self):
             self.disaggregated_network.generators_t.p.fillna(0, inplace=True)
             self.disaggregated_network.generators_t.q.fillna(0, inplace=True)
 
-            self.disaggregated_network.results = self.network.results
             log.info(
                 "Time for overall desaggregation [min]: {:.2}".format(
                     (time.time() - t) / 60
--- a/etrago/cluster/disaggregation.py
+++ b/etrago/cluster/disaggregation.py
@@ -246,16 +246,13 @@ class Disaggregation:
         # Just a simple sanity check
         # TODO: Remove when sure that disaggregation will not go insane anymore
         for line_type in line_types:
-            assert (
-                getattr(partial_network, line_type)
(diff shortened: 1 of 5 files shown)


etrago/cluster/disaggregation.py
```diff
@@ -669,15 +710,21 @@ def solve_partial_network(
...
+                            for bus_id in filtered.index
+                        }
+                    )
+                    pn_t[s].loc[:, new_columns.columns] = new_columns
 
     def transfer_results(self, *args, **kwargs):
-        kwargs["bustypes"] = ["generators", "storage_units"]
+        kwargs["bustypes"] = ["generators", "links", "storage_units", "stores"]
         kwargs["series"] = {
             "generators": {"p"},
```
ClaraBuettner, line 724: In case a pf was performed, the `q` should be also transferred for generators and storage_units. 
Could you add this?

ClaraBuettner:
> The disaggregation is currently not running, because etrago.clustering.busmap is a dictionary now (it used to be a pandas.Series). This seems to be a change in pypsa, so it is not that easy to change it directly. It works when I update a few lines in the disaggregation function. Can I push that? Or do you have a better idea to solve this problem?

I pushed my solution in 73b9e3b356ef1a6df923204d424cc7a9792e6e46 , we can also undo it when it's not fine for you.

ClaraBuettner:
I did the changes for the busmap on a sub branch of this one: [features/disaggregate-links-and-stores-update-busmap](https://github.com/openego/eTraGo/tree/features/disaggregate-links-and-stores-update-busmap)

etrago/cluster/disaggregation.py
gnn: I'd only have one question: how would I check whether a PF was performed? Or can I transfer those "results" unconditionally? The thing is, that this PR deals with `"links"` and `"stores"`, so this bug was present before updating the disaggregation. That's why I moved this to a [separate issue][0] (#644).

[0]: https://github.com/openego/eTraGo/issues/644

PR #640, opened 2023-04-17
Fix/cluster foreign gas missing buses

pieterhexen:
This fixes the problem of missing nodes in graph when not considering foreign nodes in gas clustering.

### DIFF
--- a/etrago/cluster/gas.py
+++ b/etrago/cluster/gas.py
@@ -56,6 +56,11 @@ def preprocessing(etrago):
         ch4_filter & (network_ch4.buses["country"] != "DE")
     ).sum()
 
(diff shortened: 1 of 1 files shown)


pieterhexen:
> 



> I tested it and now it works as expected. I'm not sure if I really understood why the order had to be changed, but I think it is fine and can be merged

Thanks Clara! 
The order changed because if the relevant CH4-links are added to network_ch4 after the CH4-buses have been added (like before), the links connecting the buses near Saarbrücken are neglected when `cluster_foreign_gas = False` and therefore isolating the buses. The isolated buses and then not part of the graph in the Dijkstra clustering leading to the error you observed :)

PR #568, opened 2023-01-06
Feature/#532 update to plotpy

Alonsoju95:
changes in plot.py in  calc_network_expansion to plot only DC links and in v_nom

PR #658, opened 2023-09-14
Features/improve code style

### LINKED ISSUE #445: Code style
When developing the current version of eTraGo, we followed the code style rules of `pep8`. 
Right now, it sometimes happens that the code was formatted with `black`, I think that happened because we use it in eGon-data and are used to it. 
I would like to make a decision on the code style we want to use from now on to avoid conflicts. 

I had a look at the difference between the two options: `black` seems to be more strict but also follows `pep8` rules. `pep8` sometimes allows different code style options, `black` always selects the same one. 

For me, both options would be fine. I think it will be the choice between changing less things or be more consistent to our other tools. 

@CarlosEpia @AmeliaNadal @pieterhexen @fwitte: What do you think?

ClaraBuettner:
Fixes #445 

This branch applies black, isort and flake8 changes to (nearly) all files, with the following exception:

- flake8 tests for "Can not use bare except" are currently excluded. I tried to fix those problems, but since I didn't know the error that could show up for all occurrences, I was not able to fix all of them. We should follow up on this, see #669 
- the plot.py is not checked yet, since there will be many changes in #568 
- the appl.py is not checked yet, since there are many warnings concerning the overly long comments in the args

I guess none of these things prevent us from merging this branch, I just added the limitations to not forget about them.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -143,7 +143,7 @@ args = {
         "extreme_periods": None,  # consideration of extreme timesteps; e.g. 'append'
         "how": "daily",  # type of period - only relevant for 'typical_periods'
         "storage_constraints": "soc_constraints",  # additional constraints for storages  - only relevant for 'typical_periods'
-        "n_clusters": 5,  #  number of periods - only relevant for 'typical_periods'
+        "n_clusters": 5,  # number of periods - only relevant for 'typical_periods'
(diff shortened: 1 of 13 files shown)


CarlosEpia (approved):
I tested this branch and it is working as expected.

PR #659, opened 2023-09-19
Feature/#655 country flag market model

### LINKED ISSUE #655: market_model needs a country tag
Apply buses_by_country() and geolocation_buses() also on the market model like somewhere [here](https://github.com/openego/eTraGo/blob/51240f1df63cc465a0d48746657c9a997faed265/etrago/execute/market_optimization.py#L141)

CarlosEpia:
Fixes #655

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -145,11 +145,17 @@ def geolocation_buses(self):
 
     Parameters
     ----------
-    etrago : :class:`etrago.Etrago`
+    self : :class:`etrago.Etrago`
        Transmission grid object
+       or
+       :class:'pypsa.components.Network'
 
     """
-    network = self.network
+
+    if not isinstance(self, pypsa.components.Network):
+        network = self.network
+    else:
+        network = self
 
     transborder_lines_0 = network.lines[
         network.lines["bus0"].isin(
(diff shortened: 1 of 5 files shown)


etrago/tools/utilities.py
```diff
@@ -145,11 +145,17 @@ def geolocation_buses(self):
...
+       or
+       :class:'pypsa.components.Network'
 
     """
-    network = self.network
+
+    if not isinstance(self, pypsa.components.Network):
+        network = self.network
+    else:
+        network = self
```
ClaraBuettner, line 158: Allowing different data types seems a bit dangerous, if it is possible to avoid things like this, I would do it. 
It also is a bit strange because this function is part of an object, but also works without it. 

Was there any reason why you did not change that function in a similar way as you changed the buses_by_country function? I guess this one could also operate on the network only.

etrago/tools/utilities.py
```diff
@@ -203,18 +209,19 @@ def geolocation_buses(self):
         c_bus1 = network.buses.loc[network.links.loc[link, "bus1"], "country"]
         network.links.loc[link, "country"] = "{}{}".format(c_bus0, c_bus1)
 
-    return network
+    return
 
 
-def buses_by_country(self):
+def buses_by_country(network, con):
```
ClaraBuettner, line 215: I am still wondering if it is the best option to remove functions from the object. I will think about it and reply again.

Issue #661, opened 2023-09-21
Exclusion clustering area


CarlosEpia:
Sometimes some areas of the model are more relevant than others for a specific analysis, therefore is necessary to implement somethis that keeps the highest possible resolution in a provided area whiles still clusters everything outside it.

PR #662, opened 2023-09-21
Include exclusion area in clustering

CarlosEpia:
Solves #661. The areas can be supplied by path to shapefiles or list of nuts names e.g.,  ["Cuxhaven", "Bremerhaven", "Wesermarsch", "Osterholz", "Bremen"]


etrago/appl.py
```diff
@@ -112,12 +112,15 @@
     "network_clustering_ehv": {
         "active": False,  # choose if clustering of HV buses to EHV buses is activated
         "busmap": False,  # False or path to stored busmap
+        "interest_area": False,  # False, path to shapefile or list of nuts names of not cluster area
```
birgits, line 115: I would write "of area that is excluded from the clustering"

etrago/appl.py
```diff
@@ -112,12 +112,15 @@
...
     },
     "network_clustering": {
         "active": True,  # choose if clustering is activated
         "method": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
         "n_clusters_AC": 30,  # total number of resulting AC nodes (DE+foreign)
         "cluster_foreign_AC": False,  # take foreign AC buses into account, True or False
+        "interest_area": False,  # False, path to shapefile or list of nuts names of not cluster area
```
birgits, line 122: Here, I would write "of area that is clustered to the number of clusters specified through "cluster_interest_area"". For "n_clusters_AC" above I would write an explanation, that the "interest_area" is excluded (DE+foreign-interest_area), or is it not?

etrago/appl.py
```diff
@@ -415,6 +414,13 @@ def run_etrago(args, json_path):
         a new busmap must be calculated. False or path to the busmap in csv
         format should be given.
         Default: False
+        * "interest_area": False, list, string
+        Area of especial interest that will be not clustered. It is by
+        default set to false. When an interest_area is provided, the given
+        value for n_clusters_AC will mean the total of AC buses outside the
```
birgits, line 420: The n_clusters_AC does not apply here. Also needs to be indented in order to be rendered correctly with Sphinx.

etrago/appl.py
```diff
@@ -445,13 +453,24 @@ def run_etrago(args, json_path):
             as well and included in number of clusters specified through
             ``'n_clusters_AC'``.
             Default: False.
+        * "interest_area": False, list, string
```
birgits, line 456: This needs to be adapted, as the interest area can be clustered when  "cluster_interest_area" is not False.

etrago/appl.py
```diff
@@ -445,13 +453,24 @@ def run_etrago(args, json_path):
...
+            Area of especial interest that will be not clustered. It is by
+            default set to false. When an interest_area is provided, the given
+            value for n_clusters_AC will mean the total of AC buses outside the
+            area.The area can be provided in two ways: list of
+            nuts names e.G. ["Cuxhaven", "Bremerhaven", "Bremen"] or a string
+            with a path to a shape file (.shp).
+            Default: False.
+        * "cluster_interest_area": False, int
```
birgits, line 464: I would rename this to "n_cluster_interest_area" to be more consistent with the other parameters.

etrago/appl.py
```diff
@@ -445,13 +453,24 @@ def run_etrago(args, json_path):
...
+            Number of buses to cluster all the electrical buses in the
+            exclusion area. Method provided in the arg "method" is used.
```
birgits, line 466: "interest_area" instead of exclusion area

birgits, line 466: "interest_area" instead of exclusion area

etrago/appl.py
```diff
@@ -734,12 +710,12 @@ def run_etrago(args, json_path):
 
     # snapshot clustering
     etrago.snapshot_clustering()
-
+    breakpoint()
```
birgits, line 713: Should be deleted

CarlosEpia:
Hi @ulfmueller this PR is ready for review. According to our agreement with @birgits and David, now it is possible to use also the eHV clustering while keeping the buses inside the interest area. It means that there are no 110kV lines or buses outside the provided area, therefore it is very important to define carefully the shapes since it could lead to 110kV isolated buses.

CarlosEpia:
Regarding the definition of the _Interest area_, I think that it makes sense to provide a better explanation.
- Well defined _interest areas_ should not contain dead ends in 110kV, since the eHV clustering eliminates all the connections in that voltage that are not completely contained in the _interest area_. For example, when selecting Bremen, Bremerhafen and Cuxhaven, sections of the network like the one shown below will create problems. Additionally, I would try to avoid _interest areas_ composed of geographically separated areas when using eHV.
Bremen, Bremerhafen and Cuxhaven

CarlosEpia:
As examples of well-defined areas, we can take the two shapefiles created and shared by @ulfmueller ([Fokusregion.zip](https://github.com/openego/eTraGo/files/15099572/Fokusregion.zip)), where the network topology is taken into account and no stubs can be found. Using the mentioned areas, the clustering process runs as expected.

birgits:
@CarlosEpia I added the option, to plot the nodes inside the interest area differently from the other nodes. You can remove it if it's not helpful or anything. I just added it as a suggestion as I implemented it for myself anyway.

etrago/tools/utilities.py
```diff
@@ -2973,3 +2981,260 @@ def manual_fixes_datamodel(etrago):
...
+                ), f"""'n_clusters_AC' must be greater than the number of
+                foreign countries({num_neighboring_country})"""
+
+        elec_network = network_based_on_buses(
+            etrago.network, elec_network_buses
+        )
+        area_network = network_based_on_buses(etrago.network, buses_area)
+
+    elif apply_on == "market_model":
+        elec_network_buses = etrago.network_tsa.buses[
```
birgits, line 3060: @CarlosEpia or @ClaraBuettner, I discovered that n_clusters is not assigned in this elif. Not sure if this was done in this branch or not, so maybe it's not an issue that should be fixed here.

ClaraBuettner: This function is defined in cluster.electrical.py on `dev`. In that function, n_clusters is defined for all settings and the clustering works fine. So I guess it is related to this branch. Could you have a look at it @CarlosEpia?

CarlosEpia: @birgits good catch! Thanks for the warning. Problem solved in https://github.com/openego/eTraGo/commit/cec6836f640fda2a50adb113f550452122673af5.

Issue #644, opened 2023-05-23
If a PF was performed, `q` should be considered in `transfer_results`


gnn:
At least for `"generators"` and `"storage_units"`. See also [this conversation][0] in PR #629.

[0]: https://github.com/openego/eTraGo/pull/629#discussion_r1147537303

ClaraBuettner:
done in #674

PR #674, opened 2023-10-02
Fix disaggregation

ClaraBuettner:
The original PR to disaggregate links and stores (#629) included some problems, e.g. https://github.com/openego/eTraGo/pull/629/commits/95c22f016fba3dfb22074fa7b5c0c7832a089482 assumed that `etrago.clustering.busmap `is a `pandas.Series`, but it is a dictionary. 
In addition, some commits from https://github.com/openego/eTraGo/tree/features/disaggregate-links-and-stores-update-busmap were not part of the merged branch. 
I fixed those problems in this branch. 
This branch also solves #644

### DIFF
--- a/etrago/cluster/disaggregation.py
+++ b/etrago/cluster/disaggregation.py
@@ -799,7 +799,7 @@ def update_constraints(network, externals):
 
 def run_disaggregation(self):
     log.debug("Running disaggregation.")
-    if self.clustering:
+    if self.args["network_clustering"]["active"]:
         disagg = self.args.get("disaggregation")
         skip = () if self.args["pf_post_lopf"]["active"] else ("q",)
         t = time.time()
--- a/etrago/cluster/disaggregation.py
+++ b/etrago/cluster/disaggregation.py
@@ -15,7 +15,7 @@ from etrago.tools.utilities import residual_load
 
 class Disaggregation:
     def __init__(
-        self, original_network, clustered_network, clustering, skip=()
+        self, original_network, clustered_network, busmap, skip=()
     ):
         """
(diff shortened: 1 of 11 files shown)


etrago/cluster/disaggregation.py
```diff
@@ -744,6 +755,10 @@ def solve_partial_network(
                     )
                     delta = abs((new_columns.sum(axis=1) - clt).sum())
                     epsilon = 1e-5
+                    if delta > epsilon:
+                        import pdb
+
```
CarlosEpia, line 760: I suppose you used these lines for debugging purposes. Can we exclude them now?

CarlosEpia (changes requested):
Using the default arguments but with 50 snapshots, the disaggregation shows an error:
"Sum of disaggregated time series does not match..." Do you know what could be the reason?

etrago/cluster/disaggregation.py
ClaraBuettner: I removed these lines

CarlosEpia (approved):
I tested this branch and it ran OK. It was not tested when PF post LOPF is activated.

ClaraBuettner:
The problems when disaggregating load_shedding generators after the pf_post_lopf occur because the control strategies for generators are set multiple times, and sometimes different.
That resulted in different control strategies for load_shedding generators in the aggregated and the disaggregated network. 
I think it is the best solution to set the control strategies only once, and at a central place. I will look for that, change the code and inform you when it is done.

ClaraBuettner:
> I tested this branch and it ran OK. It was not tested when PF post LOPF is activated.

I fixed the problems, it now also works when the pf was executed.

Issue #660, opened 2023-09-20
Errors in EHV clustering


birgits:
When I try to run the EHV clustering I get the following error:

```
INFO:etrago.cluster.electrical:Start ehv clustering

Busmap does not exist and will be created.

Unexpected exception formatting exception. Falling back to standard exception

Traceback (most recent call last):
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/tmp/ipykernel_10186/207841767.py", line 1, in <module>
    etrago.ehv_clustering()
  File "/home/birgit/virtualenvs/hyBit/git_repos/eTraGo/etrago/cluster/electrical.py", line 475, in ehv_clustering
    busmap = busmap_from_psql(self)
  File "/home/birgit/virtualenvs/hyBit/git_repos/eTraGo/etrago/cluster/spatial.py", line 545, in busmap_from_psql
  File "/home/birgit/virtualenvs/hyBit/git_repos/eTraGo/etrago/cluster/spatial.py", line 457, in busmap_by_shortest_path
    assert (buses[~mask].v_nom.astype(int).isin(tolvl)).all()
AssertionError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 2105, in showtraceback
    stb = self.InteractiveTB.structured_traceback(
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/ultratb.py", line 1396, in structured_traceback
    return FormattedTB.structured_traceback(
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/ultratb.py", line 1287, in structured_traceback
    return VerboseTB.structured_traceback(
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/ultratb.py", line 1140, in structured_traceback
    formatted_exception = self.format_exception_as_a_whole(etype, evalue, etb, number_of_lines_of_context,
  File "/home/birgit/virtualenvs/hyBit/lib/python3.8/site-packages/IPython/core/ultratb.py", line 1055, in format_exception_as_a_whole
    frames.append(self.format_record(record))
[...]

PR #676, opened 2023-10-05
Fixes/#660 fix ehv clustering

CarlosEpia:
Deals with #660

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -100,6 +100,7 @@ args = {
             "grid_max_abs_foreign": None,  # absolute capacity per voltage level
         },
     },
+    "delete_dispensable_ac_buses": False, # bool. Find and delete unnecesary buses
     "generator_noise": 789456,  # apply generator noise, False or seed number
     "extra_functionality": {},  # Choose function name or {}
     # Spatial Complexity:
@@ -379,6 +380,11 @@ def run_etrago(args, json_path):
             Limit overall energy production country-wise for each generator
             by carrier. Set upper/lower limit in p.u.
 
+    delete_dispensable_ac_buses: bool
+        Choose if unnecessary buses should be identified and deleted from the
+        grid. This buses have no load or generation attached. Additionally,
+        they are just connected to two other buses.
+        Default: False.
     network_clustering_ehv : bool
         Choose if you want to cluster the full HV/EHV dataset down to only the
         EHV buses. In that case, all HV buses are assigned to their closest EHV
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -103,7 +103,10 @@ args = {
     "extra_functionality": {},  # Choose function name or {}
     # Spatial Complexity:
     "delete_dispensable_ac_buses": True, # bool. Find and delete expendable buses
-    "network_clustering_ehv": False,  # clustering of HV buses to EHV buses
+    "network_clustering_ehv": {
+        "active": False,  # choose if clustering of HV buses to EHV buses is activated
+        "busmap": False, # False or path to stored busmap
+    },
     "network_clustering": {
         "active": True,  # choose if clustering is activated
         "method": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
@@ -390,11 +393,22 @@ def run_etrago(args, json_path):
         connected lines are merged. This reduces the spatial complexity without
         losing any accuracy.
         Default: True.
-    network_clustering_ehv : bool
+    network_clustering_ehv : dict
+        Choose if you want to apply an extra high voltage clustering to the
+        electrical network.
+        The provided dictionary can have the following entries:
+
+        * "active" : bool
         Choose if you want to cluster the full HV/EHV dataset down to only the
         EHV buses. In that case, all HV buses are assigned to their closest EHV
         substation, taking into account the shortest distance on power lines.
         Default: False.
+        * "busmap" : str
+        Choose if an stored busmap can be used to make the process quicker, or
+        a new busmap must be calculated. False or path to the busmap in csv
+        format should be given.
+        Default: False
+
     network_clustering : dict
         Choose if you want to apply a clustering of all network buses and
         specify settings.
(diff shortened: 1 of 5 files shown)


CarlosEpia:
Finally all the checks passed. Ready for review.

etrago/appl.py
```diff
@@ -383,6 +384,11 @@ def run_etrago(args, json_path):
             Limit overall energy production country-wise for each generator
             by carrier. Set upper/lower limit in p.u.
 
+    delete_dispensable_ac_buses: bool
+        Choose if unnecessary buses should be identified and deleted from the
+        grid. This buses have no load or generation attached. Additionally,
+        they are just connected to two other buses.
+        Default: False.
```
ClaraBuettner, line 391: I would probably not call it "unnecessary buses" that somehow sounds as if it would be a bug. 
It would be also good to mention that it affects the electrical grid.
I would suggest: 

Choose if electrical buses that are only connecting two lines should be removed.  These buses have no other components attached to them, the connected lines are merged. This reduces the spatial complexity without losing any accuracy.

etrago/appl.py
```diff
@@ -99,6 +99,7 @@
             "grid_max_abs_foreign": None,  # absolute capacity per voltage level
         },
     },
+    "delete_dispensable_ac_buses": False, # bool. Find and delete unnecesary buses
```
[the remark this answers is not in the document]
ClaraBuettner: Also you need to add the parameter to the args.json file

etrago/cluster/spatial.py
```diff
@@ -538,24 +545,26 @@ def fetch():
...
-        else:
-            cpu_cores = int(cpu_cores)
-
-        busmap_by_shortest_path(
-            etrago,
-            scn_name,
-            fromlvl=[110],
-            tolvl=[220, 380, 400, 450],
-            cpu_cores=cpu_cores,
+    if busmap:
```
ClaraBuettner, line 548: I guess you don't want to include this change

etrago/tools/utilities.py
```diff
@@ -1079,6 +1089,8 @@ def delete_dispensable_ac_buses(etrago):
     None.
 
     """
+    if etrago.args["delete_dispensable_ac_buses"] is False:
+        return
```
ClaraBuettner, line 1093: It is a bit strange that the function is called anyway. Why didn't you add the if-statement here?: https://github.com/openego/eTraGo/blob/a8515570275f68f3fcfd7a2de48becad0c689144/etrago/tools/network.py#L406

etrago/cluster/spatial.py
CarlosEpia: I included this change intentionally. Using old eHV busmaps leads to errors in the eHV due to the different arguments for the parameter "foreign_lines": {"carrier": "AC",}

etrago/tools/utilities.py
CarlosEpia: I followed the implementation of the load_shedding, with the difference that I did not include everything inside an if statement.
https://github.com/openego/eTraGo/blob/a8515570275f68f3fcfd7a2de48becad0c689144/etrago/tools/utilities.py#L586C1-L604C35

etrago/appl.py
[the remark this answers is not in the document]
CarlosEpia: I think setting this to False can save some time when using small datasets. Also setting it to True would keep the current behaviour of eTraGo. I changed it so to True, but probably both of them make perfect sense.

args.json was also updated.

etrago/appl.py
CarlosEpia: Changed.

CarlosEpia:
Comments already considered. Ready for a new review.

etrago/cluster/spatial.py
ClaraBuettner: Ok, but then I would solve it differently. Right now the busmap is only created if there is data in the table. If there is no data, you do not create the busmap and run into problems. 
And it is also not that intuitive that there is a table where busmaps are stored, but it is not used. 
I would rather change the check if the busmap from the table can be used

etrago/tools/utilities.py
ClaraBuettner: Sorry for the comment :)

etrago/cluster/spatial.py
CarlosEpia: Right now the busmap is always created. If there is one already, a message is shown to inform that it will be replaced. Since sometimes we know that the busmap can be reused, I will include a new argument called "reuse_busmap" in my next commit. It will be included in the "network_clustering_ehv" parameter.

etrago/cluster/spatial.py
```diff
@@ -476,25 +472,11 @@ def busmap_by_shortest_path(etrago, scn_name, fromlvl, tolvl, cpu_cores=4):
...
-        schema="grid",
-        if_exists="append",
-    )
+    busmap = pd.Series(df.bus1.values, index=df.bus0).to_dict()
 
-    return
+    return busmap
 
 
 def busmap_from_psql(etrago):
```
ClaraBuettner, line 482: The description and the name of the function do not represent the content anymore

CarlosEpia:
> I get this warning when I run the kmedoids-dijkstra clustering after the ehv clustering:
> 
> ```
> INFO:etrago.cluster.electrical:Network clustered to EHV-grid
> INFO:etrago.cluster.electrical:Start k-medoids Dijkstra Clustering
> INFO:etrago.cluster.electrical:There are 3 buses that have no direct
>             connection to the electric network: ['77783', '78329', '78721']
> ```
> 
> Do you have the same warning? It looks as if it should not be there...

There were 3 not electric buses being clustered based on eHV foreign buses. When the Kmedoid-Dijkstra clustering was used excluding the foreign countries, the mentioned buses were isolated. The commit [92aeaf1](https://github.com/openego/eTraGo/pull/676/commits/92aeaf1ff934e9e323441aaf778426209bb9ee9e) includes changes to find the closest eHV bus in Germany for those cases.

ClaraBuettner (approved):

PR #693, opened 2023-11-14
Fix dc_loading for line loading plot

ClaraBuettner:
This branch fixed the visualization of line_loading. Could you double-check that it works as expected?

### DIFF
--- a/etrago/tools/plot.py
+++ b/etrago/tools/plot.py
@@ -1743,7 +1743,7 @@ def plot_background_grid(network, ax, geographical_boundaries, osm):
     """
     link_widths = pd.Series(index=network.links.index, data=0)
     link_widths.loc[network.links.carrier == "DC"] = 0.3
-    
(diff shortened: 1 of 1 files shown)


CarlosEpia (approved):

CarlosEpia:
The loading plot is working correctly. I did the mentioned change regarding the problem with OSM background map.

etrago/tools/plot.py
```diff
@@ -2703,7 +2717,6 @@ def plot_grid(
             flow=flow,
             title=title,
             geomap=False,
-            boundaries=geographical_boundaries,
```
ClaraBuettner: Removing this line created some problems because the defined geographical boundaries were not used anymore. I fixed this in my latest commit.

CarlosEpia:
All the plots are working as expected. This branch can be merged into dev.

PR #703, opened 2023-12-05
Fix clustering of h2_grid nodes

ClaraBuettner:
Look for "H2_grid" carrier in network.buses. Before this change, the H2_grid nodes were not clustered at all since "H2" was never in the list of carriers


CarlosEpia (approved):
I tested this branch using eGon2035 and it works as expected.

Issue #719, opened 2024-02-06
Function "select_elec_network" can not be used for market-optimization


CarlosEpia:
The function "select_elec_network" is currently used during the market optimization and it is working well so far, but this function uses a parameter from "network_clustering" which specifies if the foreign buses should be considered or not. The  function "select_elec_network" must be able to work without dependencies from external parameters in the args.

PR #720, opened 2024-02-06
adjust function select_elec_network

CarlosEpia:
Fix a small bug when running market optimization while the parameter "cluster_foreign_AC" = True.
#719

### DIFF
--- a/etrago/cluster/electrical.py
+++ b/etrago/cluster/electrical.py
@@ -468,7 +468,7 @@ def ehv_clustering(self):
         logger.info("Network clustered to EHV-grid")
 
 
-def select_elec_network(etrago, include_foreign, apply_on="grid_model"):
+def select_elec_network(etrago, apply_on="grid_model"):
     """
     Selects the electric network based on the clustering settings specified
     in the Etrago object.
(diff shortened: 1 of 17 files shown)


etrago/cluster/electrical.py
```diff
@@ -645,14 +648,20 @@ def unify_foreign_buses(etrago):
...
     Preprocesses an Etrago object to prepare it for network clustering.
 
     Parameters
     ----------
     etrago : Etrago
         An instance of the Etrago class
+    used_for : string
+        provide information about the objective of the preprocessing. Which
+        process is going to use the result. e.g. "spatial_clustering",
+        "market_model".
```
ClaraBuettner, line 664: This parameter seems to be very similar to the other new parameter "apply_on" that is already in `dev`. 
Is the "used_for" parameter always "market_model" when applied on the market model and "spatial_clustering" if it is applied on the grid_model /etrago.network?

CarlosEpia: Yes, it is. I will adjust my PR to work using the already existing parameter "apply_on".

CarlosEpia:
Ready for a new check.

etrago/cluster/electrical.py
```diff
@@ -477,6 +477,12 @@ def select_elec_network(etrago, apply_on="grid_model"):
     ----------
     etrago : Etrago
         An instance of the Etrago class
+    apply_on: str
+        gives information about the objective of the output network. If
+        "spatial_clustering" is provided, the value assigned in the args for
```
ClaraBuettner, line 482: Is it correct that you can choose 'spatial_clustering' here? 
The code later on only allows 'grid_model' or 'market_model'

etrago/cluster/electrical.py
```diff
@@ -496,7 +502,13 @@ def select_elec_network(etrago, apply_on="grid_model"):
             """
         )
     settings = etrago.args["network_clustering"]
-    if settings["cluster_foreign_AC"]:
+
+    if apply_on == "grid_model":
+        include_foreign = settings["cluster_foreign_AC"]
+    elif apply_on == "market_model":
+        include_foreign = True
```
ClaraBuettner, line 509: There is no else statement

ClaraBuettner (approved):

Issue #699, opened 2023-11-22
Additional constraints for limiting redispatch


ClaraBuettner:
We would like to have the option to limit the redispatch (ram up and ramp down generation). There are several options on how redispatch can be limited:

- maximum (and/or minimum) overall annual redispatch, amount could be e.g. taken from NEP
- limit redispatch per market region / node

I think I will start with implementing restrictions for overall Germany.

PR #723, opened 2024-02-09
Add constraint to limit redispatch

ClaraBuettner:
Solves #699


ulfmueller (commented):
Maybe a stupid question: does it make sense to some how also include the ramp_down redispatch?

ulfmueller (approved):

PR #732, opened 2024-03-05
Add try/except statement to allow importing etrago object before optimization when "market-grid" is selected

ClaraBuettner:
Just a small change that allows to import the object when the market-grid optimization is selected but not yet executed


CarlosEpia (approved):
Looks good to me

PR #749, opened 2024-06-25
call fuction get_switchable_as_dense before p_max_pu is modified

### LINKED ISSUE #748: ramp_up generators not dispatching
The ramp_up generators are not dispatching. This leads to unfeasibilities and mandatory use of load_shedding. 
- It was found that the `p_max_pu` values for all the generators are 0.
- `p_min_pu` are well defined and correspond to the negative value of `p` from the market optimization.

CarlosEpia:
This PR fixes #748.

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -731,15 +731,15 @@ def run_etrago(args, json_path):
     # conduct lopf with full complex timeseries for dispatch disaggregation
     etrago.temporal_disaggregation()
 
-    # # start power flow based on lopf results
-    # etrago.pf_post_lopf()
+    # start power flow based on lopf results
+    etrago.pf_post_lopf()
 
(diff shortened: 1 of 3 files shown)


etrago/appl.py
```diff
@@ -729,17 +729,17 @@ def run_etrago(args, json_path):
     etrago.optimize()
 
     # conduct lopf with full complex timeseries for dispatch disaggregation
-    etrago.temporal_disaggregation()
+    # etrago.temporal_disaggregation()
```
ClaraBuettner, line 732: These changes look somehow like they should not be part of dev... Or was there a reason for changing it?

CarlosEpia: Agree. The args were restored.

ClaraBuettner (approved):

PR #714, opened 2024-01-30
Update module structure

ClaraBuettner:
This branch slightly adjusts the structure of etrago. There are now more submodules, and especially the tools module is now smaller. 
The different modules are quite similar to the flow-chart we often use to show the functionalities of eTraGo, which makes it easier to combine both. The structure now looks like this: 

- etrago
- etrago.appl
- etrago.network
- etrago.analyze
-- etrago.analyze.calc_results
-- etrago.analyze.plot
- etrago.cluster
-- etrago.cluster.electrical
-- etrago.cluster.gas
-- etrago.cluster.snaphot
-- etrago.cluster.spatial
- etrago.disaggregate
-- etrago.disaggregate.spatial
-- etrago.disaggregate.temporal
-  etrago.execute
-- etrago.execute.grid_optimization
-- etrago.execute.market_optimization
-- etrago.execute.sclopf will be added
-  etrago.tools
-- etrago.tools.io
--- etrago.tools.io.db
-- etrago.tools.constraints
-- etrago.tools.extendable
-- etrago.tools.utilities 

I am not sure if that is the very best structure but it is in my opinion already an improvement to the modules before, where many things were in the tools module. It might make sense to move e.g. constraints and extendable to the execute module. We can discuss the structure. This setting is already working, the tool itself and als the rtd is working fine.

PR #752, opened 2024-08-08
Features/update module structure v2

ClaraBuettner:
I created a new branch and PR for the new structure because the previous one (#714) has so many merge conflicts. 

I ran a small test which was working as expected. The rtd is also working fine.


KathiEsterl:
Looks fine to me! Here some comments:

1. To be very consistent, we could also change cluster.snapshot to cluster.temporal. Seems to be optional for me. What do you think?
2. In the rtd, all folders are marked as packages, only etrago.analyze is marked as module although it also is a folder. I am not sure whey this is the case, what it means exactly and whether it is important.

I think testing the following settings should cover the changes:

- [ ] check if clustering (spatial and temporal) are working
- [ ] check if additional constrains are working
- [ ] check if calc_results is working
- [ ] check some plotting functions
- [ ] check if disaggegation is working

ClaraBuettner:
> Looks fine to me! Here some comments:
> 
>     1. To be very consistent, we could also change cluster.snapshot to cluster.temporal. Seems to be optional for me. What do you think?
> 
>     2. In the rtd, all folders are marked as packages, only etrago.analyze is marked as module although it also is a folder. I am not sure whey this is the case, what it means exactly and whether it is important.
> 
> 
> I think testing the following settings should cover the changes:
> 
>     * [x]  check if clustering (spatial and temporal) are working
> 
>     * [x]  check if additional constrains are working
> 
>     * [x]  check if calc_results is working
> 
>     * [x]  check some plotting functions
> 
>     * [x]  check if disaggregation is working

I renamed the section in the read the docs for etrago.analyze.

KathiEsterl (approved):

Issue #762, opened 2024-10-12
Too many deprecation warnings


CarlosEpia:
Using the current dev version, several deprecation warnings make it difficult to filter the real important messages.

CarlosEpia:
Solved via #763

PR #763, opened 2024-10-12
Features/#762 too many deprecation warnings

CarlosEpia:
This PR solves the issue #762. It can be merged once all these different deprecation warnings are addressed:
- [x] /home/carlos/env/etrago_624/lib/python3.10/site-packages/sklearn/base.py:486: UserWarning: X has feature names, but KMeans was fitted without feature names
  warnings.warn(
- [x] /home/carlos/env/etrago_624/lib/python3.10/site-packages/pypsa/clustering/spatial.py:255: FutureWarning: The provided callable <function min at 0x70a2744d0ee0> is currently using SeriesGroupBy.min. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string "min" instead.
  aggregated = df.groupby(grouper).agg(static_strategies)
- [x] /home/carlos/git/eTraGo/etrago/cluster/spatial.py:278: FutureWarning: DataFrame.groupby with axis=1 is deprecated. Do `frame.T.groupby(...)` without axis instead.
  pnl_df = df_agg.groupby(grouper, axis=1).sum()
- [x] /home/carlos/git/eTraGo/etrago/tools/utilities.py:290: FutureWarning: The geopandas.dataset module is deprecated and will be removed in GeoPandas 1.0. You can get the original 'naturalearth_lowres' data from https://www.naturalearthdata.com/downloads/110m-cultural-vectors/.
  path = gpd.datasets.get_path("naturalearth_lowres")


etrago/tools/utilities.py
```diff
@@ -287,9 +288,10 @@ def buses_by_country(self, apply_on="grid_model"):
     con = self.engine
     germany_sh = gpd.read_postgis(query, con, geom_col="geometry")
 
-    path = gpd.datasets.get_path("naturalearth_lowres")
-    shapes = gpd.read_file(path)
-    shapes = shapes[shapes.name.isin([*countries])].set_index(keys="name")
+    # read Europe borders. Original data downloaded from naturalearthdata.com/
+    # under Public Domain license
+    path = Path(".") / "tools" / "shapes_europe"
```
ClaraBuettner, line 293: In general, we shouldn't add data to the github repository, esp. things like shapefiles that are not that small.   Was there no other option than adding it to the repo?

CarlosEpia: It is also possible to download the shapes during the execution, but since the 5 files are in total 38,5 kB (and no licence is required) I think it is a good option to have it here. Also, it has the advantage that no internet connection would be needed and it would not be susceptible to changes in the webpage. If you think that it is a better idea to access the file from the webpage I could make those changes.

ClaraBuettner: The file sizes is not that big, but it adds up for each branch you have. And when there is no license needed to use it, it doesn't necessarily mean that we are allowed to republish it. So I guess we would need to check that as well. 
The release notes of the geopandas version that added the deprecation warning refers to the "geodatasets" tool. Have you had a look at this?

CarlosEpia: I deleted the files and included instructions to download them when necessary in fc5c4a81c91a71b3a24ab43d780ac0e203ff566d and b85dcd77b0af16a1d3658c5236330e5a8146ebe2

etrago/tools/utilities.py
```diff
@@ -287,9 +290,24 @@ def buses_by_country(self, apply_on="grid_model"):
     con = self.engine
     germany_sh = gpd.read_postgis(query, con, geom_col="geometry")
 
-    path = gpd.datasets.get_path("naturalearth_lowres")
-    shapes = gpd.read_file(path)
-    shapes = shapes[shapes.name.isin([*countries])].set_index(keys="name")
+    # read Europe borders. Original data downloaded from naturalearthdata.com/
+    # under Public Domain license
+    path_countries = Path(".") / "tools" / "shapes_europe"
```
ClaraBuettner, line 295: There is already a sub-directory called "data" in etrago. I would either store the file in that folder or in the working directory.

CarlosEpia: Sure. The path was changed in 712eda0ba489fc31c33dcd7f5cf4449abffa1609.

ClaraBuettner (approved):

PR #765, opened 2024-10-23
Do not overwrite clustering ans busmap object when creating the marke…

### LINKED ISSUE #764: Store correct busmap when using the consecutive method
When the consecutive method is executed, the busmap stored in etrago.busmap refers to the clustering to the market zones. This is a bit misleading since we mainly need the busmap for the grid.

ClaraBuettner:
…t_model

Fixes #764 

@CarlosEpia Could you have a look if I made the change you recommended? 
I already tried it out and it seems to work.

### DIFF
--- a/etrago/execute/market_optimization.py
+++ b/etrago/execute/market_optimization.py
@@ -311,7 +311,6 @@ def build_market_model(self):
         apply_on="market_model",
     )
 
-
     net = clustering.network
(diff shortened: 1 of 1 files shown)


CarlosEpia (approved):
I tested this branch and the described problem is solved. It can be merged.

PR #769, opened 2024-11-04
Fixes #768 fix positions of foreign clustered buses in market_model

### LINKED ISSUE #768: Two buses for DE when using market optimization
When using the market optimization we are getting 2 buses in Germany (expected 1). The problem appears to happen in the very last part of build_market_model(), where the bus corresponding to AT is assigned to DE.
https://github.com/openego/eTraGo/blob/4c9d46ce7cb6ece349ff34784d03e12be2ff09c8/etrago/execute/market_optimization.py#L408-L409

Initially, there are plenty of buses in Austria, but the resulting clustered bus is located inside Germany. I Think it is due to the average x and y values of the original positions (see picture below).
image
A clean solution could be to use the parameter _medoid_idx_  to fix the position of the resulting clustered buses.
https://github.com/openego/eTraGo/blob/4c9d46ce7cb6ece349ff34784d03e12be2ff09c8/etrago/execute/market_optimization.py#L304-L312

CarlosEpia:
Fixes #768

### DIFF
--- a/etrago/execute/market_optimization.py
+++ b/etrago/execute/market_optimization.py
@@ -312,6 +312,18 @@ def build_market_model(self):
     )
 
     net = clustering.network
+
+    # Adjust positions foreign buses
+    foreign = self.network.buses[self.network.buses.country != "DE"].copy()
+    foreign = foreign[foreign.index.isin(self.network.loads.bus)]
+    foreign = foreign.drop_duplicates(subset="country")
+    foreign = foreign.set_index("country")
+
+    for country in foreign.index:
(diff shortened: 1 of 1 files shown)


ClaraBuettner:
The code looks good to me. Do I need to try it out or did you already test it in depth?

CarlosEpia:
I tested it using eHV and Dijsktra (separately) and the results look as expected. You could give it a try using the default parameters + eHV clustering just to be sure.

ClaraBuettner (approved):
I finally tested what you suggested and it was running successful.

Issue #770, opened 2024-11-08
Negative load_shedding for debugging purposes


CarlosEpia:
When optimizing scenarios with low/no expansion allowed and fixed generation, negative load_shedding is useful to facilitate the debugging process when the optimization is infeasible.

ClaraBuettner:
This issue was solved, right  @CarlosEpia?

CarlosEpia:
Right. I'll go ahead and close this issue.

PR #771, opened 2024-11-08
Features/#770 negative load shedding for debugging purposes

CarlosEpia:
Deals with #770

### DIFF
--- a/etrago/tools/utilities.py
+++ b/etrago/tools/utilities.py
@@ -643,7 +643,12 @@ def connected_transformer(network, busids):
     return network.transformers[mask]
 
 
-def load_shedding(self, temporal_disaggregation=False, **kwargs):
+def load_shedding(
+    self,
+    temporal_disaggregation=False,
+    negative_load_shedding=True,
+    **kwargs,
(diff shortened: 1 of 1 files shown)


etrago/tools/utilities.py
```diff
@@ -702,6 +707,31 @@ def load_shedding(self, temporal_disaggregation=False, **kwargs):
             "Generator",
         )
 
+        if negative_load_shedding:
+            # Add negative load shedding generators for AC buses
+            ac_buses = network.buses[network.buses.carrier == "AC"].index
```
ClaraBuettner, line 712: This way the negative load shedding can not solve the problems for the EV buses. Could you add the EV buses here?

CarlosEpia: Right. I made some changes in 97642abbea11b3e99a3cc425e81099730dd61e8e to make the allocation of the negative load shedding carrier flexible.

ClaraBuettner (approved):
I did not test it since I guess you already did that. The code looks fine, so I approve this PR.

PR #775, opened 2025-01-14
Features/#766 develop h2 grid clustering

### LINKED ISSUE #766: Develop H2 grid clustering
As our data models will have a H2 grid very soonish, I think we need to adjust the H2 grid clustering. So far the H2 buses are automatically clustered depending on the number of CH4 buses (see [here](https://github.com/openego/eTraGo/blob/4c9d46ce7cb6ece349ff34784d03e12be2ff09c8/etrago/cluster/gas.py#L272)). This could even stay like it (although we might want to have different ch4 to h2 grid resolutions (another question!)) - but what we definitely need is some code which aggregates the new H2 lines. As I see it the [gas_postprocessing function ](https://github.com/openego/eTraGo/blob/4c9d46ce7cb6ece349ff34784d03e12be2ff09c8/etrago/cluster/gas.py#L307) has to be also executed and slightly (hopefully) adjusted for the h2 grid and a corresponding busmap. 
Who should tackle this? @CarlosEpia @lenzim97 
@CarlosEpia you probably have the most knowledge here, right?

ClaraBuettner:
Fixes #766


CarlosEpia:
So far, this branch is not dealing with buses with carrier "H2" or "H2_saltcavern".

CarlosEpia:
Since @ClaraBuettner can not be assigned like reviewer, is it OK for you @lenzim97 to review and approve this PR?

ClaraBuettner:
Sorry, I thought I can create a PR and ask myself to review it. I can still do this and either accept that I just leave a comment as an approval or we close this PR and @CarlosEpia creates a new one I can review "officially"

CarlosEpia:
I like the first option better. @ClaraBuettner it is ok for me to receive the approval via comment and merge it when you say it.

etrago/cluster/electrical.py
```diff
@@ -166,7 +166,7 @@ def find_de_closest(network, bus_ne):
         "H2_saltcavern": "power_to_H2",
         "dsm": "dsm",
         "Li ion": "BEV charger",
-        "Li_ion": "BEV_charger",
```
ClaraBuettner, line 169: Li_ion is the carrier that is used in eGon2035 (and we also just changed it in eGon100RE), so it should not be deleted here

etrago/cluster/electrical.py
```diff
@@ -166,7 +166,7 @@ def find_de_closest(network, bus_ne):
...
+        "O2": "power_to_O2",
```
ClaraBuettner, line 169: Did you agree on this with @lenzim97? 
As far as I know there will be no "power_to_O2" link anymore.

etrago/cluster/gas.py
```diff
@@ -960,14 +977,24 @@ def run_spatial_clustering_gas(self):
...
 
         if settings["active"]:
             method = settings["method_gas"]
             logger.info(f"Start {method} clustering GAS")
 
-            gas_network, weight, n_clusters = preprocessing(self)
+            ch4_network, weight_ch4, n_clusters_ch4 = preprocessing(
+                self, "CH4"
+            )
+            if scn not in ["eGon2035", "status2019"]:
```
ClaraBuettner, line 994: Do you think there is a more generic way for this, like checking if there are H2-pipelines? 
If it stays like this we have to check for each new scenario if it is correct.

etrago/appl.py
```diff
@@ -124,6 +124,7 @@
         "cluster_foreign_AC": False,  # take foreign AC buses into account, True or False
         "method_gas": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
         "n_clusters_gas": 14,  # total number of resulting CH4 nodes (DE+foreign)
+        "n_clusters_h2": 14,  # total number of resulting H2 nodes (DE+foreign)
```
ClaraBuettner, line 127: This needs to be changed in the args.json as well

etrago/appl.py
```diff
@@ -124,6 +124,7 @@
         "cluster_foreign_AC": False,  # take foreign AC buses into account, True or False
         "method_gas": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
         "n_clusters_gas": 14,  # total number of resulting CH4 nodes (DE+foreign)
```
ClaraBuettner, line 126: I always need at least 15 gas and H2 buses. Is that correct? If yes, I would change the default value here

etrago/cluster/electrical.py
lenzim97: Yes correct, i changed the carrier for the O2-links to "PtH2_O2", so there is no more "power_to_O2"-links

ClaraBuettner: I am still not completely sure if that is the right place for adding the O2 buses and links to the clustering. 
@CarlosEpia: When I remember correctly, the list above is related to components that should be clustered considering the AC nodes and their new clusters, right? 
But the O2-components are rather connected to the electrolyzers, which are not always clustered in this part if I am not mistaken. 
I am mainly wondering because I get a huge warning when I run the clustering that the ~1400 O2-buses have no direct connection to the AC buses.

ClaraBuettner:
I am afraid, we need some more changes in the clustering for the 100%-scenario since some connections were added or changed. 
But since this is not only connected to the clustering of the H2 grid, I will create a new issue for that. 
That way, we can hopefully merge this one soon :)

CarlosEpia:
Sure. I will apply the changes regarding your comments and we can proceed with another issue to deal with the new changes in the structure of the model.

etrago/cluster/gas.py
CarlosEpia: Change done in a9bdfe01c2e285637fb9d539d3d484f4e4d98b45

etrago/appl.py
CarlosEpia: Changed in 2194e60ba83855ce305a0d886e27e39f62242913

etrago/appl.py
CarlosEpia: Changed in 2194e60ba83855ce305a0d886e27e39f62242913

CarlosEpia:
@ClaraBuettner I already apply the required changes. As you mentioned, some extra changes are required to have this clustering working with the last version of eGon100RE

etrago/cluster/electrical.py
CarlosEpia: acf4ca46f58164ed5d17989ce1177a30c5373fde

etrago/cluster/electrical.py
CarlosEpia: This change was discarded in acf4ca46f58164ed5d17989ce1177a30c5373fde to be done in issue #782

PR #838, opened 2026-02-11
Update temporal disaggregation

### LINKED ISSUE #624: improve temporal disaggregation
The temporal disaggregation can be improved by calculating the slices (in which the optimization problem is divided into to limit RAM usage) in parallel.

ClaraBuettner:
The temporal disaggregation uses the same rolling horizon approach as the market optimization on this branch. 
It works with pyomo and linopy.

Fixes #624


etrago/execute/__init__.py
KathiEsterl, line 573: Soll das pdb für den Fehlerfall hier bleiben oder ist das irgendwann mal übrig geblieben?

ClaraBuettner: It was once the idea to keep it in there in case only one slice fails. 
So it was meant to stay there. But we could also remove it if it creates confusion

KathiEsterl: No, we can also keep it! It might make sense. Just wasnt sure whe I found it.

KathiEsterl: No, we can also keep it! It might make sense. Just wasnt sure whe I found it.

KathiEsterl:
I reviewed the marked changes and also examined the code more thoroughly using pdb. Everything seems to make sense for me and the interrim-results I looked into were as expected.

For the final approval, I suggest waiting for the results of the larger calculations you have already initiated, particularly those incorporating the additional cross_border_flow_per_country functionality.

Is there anything else I should do? :)

ClaraBuettner:
I think there is nothing else to do. If you like, you could already take a look at the results from a bigger calculation without the cross_border_flow constraint, but it is also not that urgent :)

KathiEsterl (approved):

PR #851, opened 2026-04-28
Features/#388 improve aggregation strategies

### LINKED ISSUE #388: Strategy to cluster efficiency
During a meeting concerning the disaggregation the question popped up if it would make sense to take instead of the arithmetic mean a p_nom wheighted mean for the efficiency of links. See here: https://github.com/openego/eTraGo/blob/dev/etrago/cluster/networkclustering.py#L275
Nothing too important, just s.th. we can decide on while developing the disaggregation method. Maybe there are other aggregation method which can be rethought as well. 
If you have any optinions on this already @ClaraBuettner @CarlosEpia. Feel free to share them.

KathiEsterl:
Closes #388

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -124,7 +124,7 @@ args = {
             "algorithm": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
             "remove_stubs": False,  # remove stubs before kmeans clustering
             "use_reduced_coordinates": False,  # if True, do not average cluster coordinates (in remove stubs)
-            "line_length_factor": 1,  # Factor to multiply distance between new buses for new line lengths
+            "line_length_factor": 1.15,  # Factor to multiply distance between new buses for new line lengths
             "random_state": 42,  # random state for replicability of clustering results
             "n_init": 10,  # affects clustering algorithm, only change when neccesary
             "max_iter": 100,  # affects clustering algorithm, only change when neccesary
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -446,7 +446,7 @@ def run_etrago(args, json_path):
             * "line_length_factor" : float
                 Defines the factor to multiply the crow-flies distance
                 between new buses by, in order to get new line lengths.
-                Default: 1.
+                Default: 1.15.
             * "random_state" : int
                 Random state for replicability of clustering results. Default: 42.
             * "n_init" : int
--- a/etrago/cluster/gas.py
+++ b/etrago/cluster/gas.py
@@ -42,7 +42,6 @@ if "READTHEDOCS" not in os.environ:
         group_links,
         kmedoids_dijkstra_clustering,
         strategies_buses,
-        strategies_generators,
         strategies_one_ports,
     )
     from etrago.tools.utilities import set_control_strategies
@@ -421,7 +420,13 @@ def sector_coupled_clustering_strategy(etrago):
     return strategy
 
 
-def gas_postprocessing(etrago, busmap, medoid_idx=None, apply_on="grid_model"):
+def gas_postprocessing(
+    etrago,
+    busmap,
+    medoid_idx=None,
+    apply_on="grid_model",
+    aggregate_generators_carriers=None,
+):
     """
     Performs the postprocessing for the gas grid clustering based on the
     provided busmap
@@ -530,17 +535,11 @@ def gas_postprocessing(etrago, busmap, medoid_idx=None, apply_on="grid_model"):
     busmap = busmap.astype(str)
     busmap.index = busmap.index.astype(str)
 
-    if apply_on == "market_model":
-        aggregate_generators_carriers = []
-    else:
-        aggregate_generators_carriers = None
-
     network_gasgrid_c = get_clustering_from_busmap(
         network,
         busmap,
-        aggregate_generators_carriers=aggregate_generators_carriers,
+        aggregate_generators_carriers,
         one_port_strategies=strategies_one_ports(),
-        generator_strategies=strategies_generators(),
         bus_strategies=strategies_buses(),
     )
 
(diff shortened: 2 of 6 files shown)


etrago/cluster/spatial.py
```diff
@@ -67,7 +67,11 @@
     "pieterhexen, fwitte, AmeliaNadal, cjbernal071421"
 )
 
-# TODO: Workaround because of agg
+
+# Aggregation strategies:
+# PyPSA provides default strategies, often capacity-weighted strategies.
+# This is only appropriate for non-extendable components.
+# Therefore, we define and use custom strategies instead.
```
ClaraBuettner: Does that mean that capacity-weighted strategies are used for all non-extendable components?

KathiEsterl:
I only added the capacity-weighted aggregation for generators as those are the only components ifor which it makes sense in our scenarios. I also added a check for the case someone starts extending generator capacities as part of the optimization. One special case are sme time-depedant parameters of links. 

I wrote a short explanation in the issue #388

ClaraBuettner:
Ah okay, so it is not possible to do it in a different way for e.g. extendable links and non-extendable links? I thought that this would be the idea, but I might have also just misunderstood this.

KathiEsterl:
I thought about this as well, but I wasn’t sure whether we actually want that, since it would mean applying different aggregation strategies across scenarios and scenario variations. Given that we are primarily comparing lowflex and fullflex variations, I thought that might not be desirable.

KathiEsterl:
After a brief in-person discussion (not documented here), I’ll take another look at this, in particular the handling of initial capacities of 0. In principle, these should not distort the aggregation. If that holds, we could continue using a capacity-weighted approach and, where available, use the initial capacity as the weighting factor.

KathiEsterl:
Change of plans after a renewed review:

We decided to consistently use the capacity-weighted aggregation approach wherever it makes sense. This is now explicitly documented and applied throughout the implementation, even in cases where it already matches the default behavior in PyPSA, to ensure long-term transparency and traceability.

We also decided that it is acceptable for expandable components to use existing capacities as aggregation weights, as described and documented accordingly.

In addition, we verified that the approach behaves correctly in edge cases, including components with zero initial capacities.

All relevant occurrences were reviewed accordingly.

etrago/appl.py
```diff
@@ -124,7 +124,7 @@
             "algorithm": "kmedoids-dijkstra",  # choose clustering method: kmeans or kmedoids-dijkstra
             "remove_stubs": False,  # remove stubs before kmeans clustering
             "use_reduced_coordinates": False,  # if True, do not average cluster coordinates (in remove stubs)
-            "line_length_factor": 1,  # Factor to multiply distance between new buses for new line lengths
+            "line_length_factor": 1.15,  # Factor to multiply distance between new buses for new line lengths
```
ClaraBuettner, line 127: Does it make sense to change it accordingly in the args.json?

etrago/cluster/gas.py
```diff
@@ -530,17 +535,11 @@ def gas_postprocessing(etrago, busmap, medoid_idx=None, apply_on="grid_model"):
...
-    else:
-        aggregate_generators_carriers = None
-
     network_gasgrid_c = get_clustering_from_busmap(
         network,
         busmap,
-        aggregate_generators_carriers=aggregate_generators_carriers,
+        aggregate_generators_carriers,
         one_port_strategies=strategies_one_ports(),
-        generator_strategies=strategies_generators(),
```
ClaraBuettner, line 543: Is there a reason why these strategies are not used anymore?

etrago/cluster/spatial.py
```diff
@@ -239,6 +242,12 @@ def normed_or_uniform(x):
             else pd.Series(1.0 / len(x), x.index)
         )
 
+    def align_strategies(strategies, keys):  # component nicht mehr nötig
```
ClaraBuettner: There is still a comment in German here :)

etrago/cluster/spatial.py
```diff
@@ -239,6 +242,12 @@ def normed_or_uniform(x):
...
+        strategies |= {
+            k: _make_consense_links for k in set(keys).difference(strategies)
+        }
+        return {k: strategies[k] for k in keys}
```
ClaraBuettner, line 244: What is the purpose of this function? Does is set the strategy to "_make_consense_links" for all attributes that are not already in the strategies?

etrago/cluster/spatial.py
```diff
@@ -265,15 +274,22 @@ def arrange_dc_bus0_bus1(network):
...
 
-    weighting = links.p_nom.groupby(grouper, axis=0).transform(
-        normed_or_uniform
-    )
     strategies = strategies_links()
     strategies.update(cus_strateg)
-    strategies.pop("topo")
-    strategies.pop("geom")
+    # strategies.pop("topo")
+    # strategies.pop("geom")
```
ClaraBuettner: Does it make sense to remove these lines instead of commenting them?

etrago/appl.py
KathiEsterl: done

etrago/cluster/gas.py
KathiEsterl: changed that one back

etrago/cluster/spatial.py
KathiEsterl: deleted

etrago/cluster/spatial.py
KathiEsterl: added a desciption, it is copied from PyPSA

etrago/cluster/spatial.py
KathiEsterl: deleted

ClaraBuettner (approved):

PR #853, opened 2026-05-19
Add constraints that replace small values in RHS with zero

### LINKED ISSUE #844: Deal with very small values to avoid numerical problems
Our optimization model tends to have very large coefficient ranges.
Gurobi already often warns that this could cause numerical problems. The model statistics look e.g. like this: 

```
Coefficient statistics:
  Matrix range     [4e-03, 3e+02]
  Objective range  [4e-04, 5e+04]
  Bounds range     [7e+07, 7e+07]
  RHS range        [6e-25, 2e+08]
```

In this example, the RHS are the main problem (which is often the case). These include e.g. possible dispatch timeseries (p_max_pu * p_nom) and p_set of loads.
The range might be even bigger as soon as we introduce focus regions which have a higher spatial resolution. 

We should try to replace very small values (under the FeasibilityTolerance) with 0, to lower the RHS range. The impact on the results would be negligible, esp. when it is in the tolerance anyhow. 

I already did some tryouts in commit https://github.com/openego/eTraGo/pull/836/changes/c8050e85e4dc5ddf68975223bff80b6c9d823015, but it somehow created more problems, so I reverted it again.

ClaraBuettner:
Fixes #844 

A small test (including temporal desaggregation) worked well. I still have to try it out in a bigger calculation. 
When that is done, the PR can be reviewed and merged


ClaraBuettner:
The test run with a bigger calculation was successful (including the last changes), so I think this is finally ready for review

KathiEsterl (approved):

Issue #856, opened 2026-05-21
Cleanup and Restructure of Exported Results and Stored Networks


KathiEsterl:
At the moment, several intermediate results are exported multiple times, sometimes with misleading naming/descriptions, creating a risk of confusion with final results.
The handling of networks within the etrago object also needs cleanup and clearer semantics.

**Proposed restructuring of the results export:**

results/
├── README.md
├── args.yml (or settings.yml?)
├── spatial_aggregation.json (busmap)
├── original_grid_topology/
├── premarket_optimization/
├── market_optimization/
├── grid_optimization/
├── temporal_disaggregation/
├── non_linear_powerflow/
└── disaggregation/

- README.md: gives an overview of all exported files and folders, explains which step produced which outputs.
- args.yml: stores the calculation settings corresponding to the run, alternative naming could be discussed if more suitable.
- spatial_aggregation.json: Explicitly documents the spatial clustering mapping.
- original_grid_topology/: contains the original grid topology export with one snapshot (former: clustering/)
- premarket_optimization/
- market_optimization/
- grid_optimization/
- temporal_disaggregation/
- non_linear_powerflow/
- disaggregation/

Folders should only contain PyPSA network CSV exports.
Naming and semantics should clearly distinguish intermediate from final results.

**Additional TODOs:**
- etrago.network should always reference the latest/current network state.
- etrago.market_model and etrago.grid_opti_model should remain available for final result comparison.
- The import function should automatically determine and load the “most final” available network based on the calculation workflow/order above.
- Evaluate whether NetCDF (.nc) files should be used instead of CSV table exports.

ClaraBuettner:
Two thoughts on this:
- With the current implementation, a the overall result folder is > 80GB. So it would be good to find some ways to reduce it.
I tried it with one disaggregated_network: When using csv files, the size is about 10GB, when using netcdf, it is 6GB. The export is also quicker. So I guess it makes sense to switch to netcdf. 
- We should take into account that eGo currently uses etrago.disaggregated_network. If we change where the disaggregated network in etrago is stored, we need to change it in eGo accordingly.

Issue #811, opened 2025-10-08
Improve structure and implementation of spatial clustering


KathiEsterl:
**Description**

The current implementation of spatial clustering has grown historically and has become difficult to understand and maintain.
Functions related to spatial clustering are spread across different modules and sectors within the cluster/ folder without a clear structure or consistent logic.

This issue aims to reorganize and improve both the structure and the efficiency of the spatial clustering code.

**Goals**

- Refactor the code structure in cluster/ related to spatial clustering for better readability and maintainability.

- Improve computational efficiency and reduce redundancy.

- Clearly separate and document the following aspects: Arguments and settings, Imported/exported data, Pre- and postprocessing steps (especially regarding foreign countries)

**Notes**

Work was initially started in issue #800

KathiEsterl:
Also see #761

KathiEsterl:
more notes / ToDos:

- [ ] check creation and re-creation of busmaps
- [ ] special focus also on ehv clustering
- [x] check handling of foreign countries
- [x] weight-csv etc.

KathiEsterl:
Another question: Should we update the technologies considered when weighting the buses in preparation of the spatial clustering? e.g. heat pumps and e-mobility loads or power generators modelled as links like chps, gas turbines etc.

KathiEsterl:
As part of #800, I removed the option to save and load csv-files for the weighting of the clustering as this option is rather redundant with the option of loading defned busmaps in the first place. 

However, this option of using busmaps does not work at the moment, at least for the gas clustring. This needs to be fixed.
As part of this, the question came up, why we have several busmaps for the different sectors and not only one busmap for all sectors.

KathiEsterl:
> more notes / ToDos:
> 
>     * [ ]  check creation and re-creation of busmaps[x]  check handling of foreign countries[x]  weight-csv etc.

Note to myself:

It looks like 'unify_foreign_buses', as part of the preprocessing, conducts a clustering of foreign buses in any case (regardless of the settings 'per_country') and within the postprocessing afterwards overwrites the corresponding results or not depending on the settings 'per_country'. This does not seem to be very efficient.

Issue #761, opened 2024-09-25
Gas, sector coupling and electrical clustering can not be run independently


CarlosEpia:
When using spatial clustering, the argument args["network_clustering"]["active"] decides if the electrical AND gas clustering are activated. Therefore it is not possible to use one of them independently. In addition, the sector clustering has its own "active" parameter in the args, but it is only executed when the gas clustering is used.

There is a dirty provisional solution for using gas clustering alone. Set args["network_clustering"]["active"] = False initially and change it after the electrical clustering is skipped and before calling the gas clustering:

etrago.spatial_clustering()
**args["network_clustering"]["active"] = True**
etrago.spatial_clustering_gas()

KathiEsterl:
I started tackling this issue when starting to work on Issue #800.

You can see my changes here: b9d7eaaa1b0b89072951a60c33660f5985edb5ff

**Open ToDos**: 

- [ ] Check if only gas_clustering without AC_clustering runs correctly (it is running through, but I haven't checked if the results are plausible; there are quite many warnings in the gas_clustering)
- [ ] Separate the sector_coupled_clustering from the gas_clustering
I would propose to add a new file cluster/other_sectors.py and shift the corresponding code. I would then also combine the electrical clustering, the gas clustering and the sector coupled clustering in one function etrago.spatial_clustering() (so remove etrago.gas_clustering()). Additionally, as soon as this is done, within the args the setting for the sector_coupled_clustering needs to be pulled out from ["network_clustering"["gas_grids"] to an own sub-category to make this independence clear. 
- [ ] Separate the h2 from the ch4 clustering and pot. add new args like h2-busmap or h2-wight-csvs

PR #863, opened 2026-07-01
Features/#856 cleanup results folder

ClaraBuettner:
Related to #856

### DIFF
--- a/etrago/appl.py
+++ b/etrago/appl.py
@@ -89,7 +89,7 @@ args = {
     "scn_extension": None,  # None or array of extension scenarios
     # Export options:
     "lpfile": False,  # save pyomo's lp file: False or /path/to/lpfile.lp
-    "csv_export": "results",  # save results as csv: False or /path/tofolder
+    "export_results_path": "results",  # save results as csv: False or /path/tofolder
     # Settings:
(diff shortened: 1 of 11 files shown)


KathiEsterl:
I checked the code and performed several local test runs using different settings. I also exported and imported both the new and the old structured results.

I made a few minor changes:

- Fixed the incorrect path argument for the market optimization.
- Added the new path argument to args.json.
- Renamed the pf_post_lopf folder while maintaining backward compatibility, so both new and existing structured results can still be imported.
- Removed the export of sector-specific busmaps as we already have a general one for all buses and can anyhow not use the other one at the moment.

@ClaraBuettner Maybe you can have a look into the changes I did. If you are fine with them, we can merge - I would say :)

KathiEsterl:
Additionally, I removed the option to load sector-specific busmaps via the arguments and applied a few quick fixes to ensure everything continues to work, as this option is currently not functioning correctly. This will be addressed as part of issues #811 and #761. However, the implementation will ultimately use a single general busmap covering all sectors instead.

Ready for your re-review, @ClaraBuettner :)

KathiEsterl (approved):

Issue #843, opened 2026-03-10
Release 0.11.0


KathiEsterl:
- [x] tag all issues that should be part of the release
- [x] check if all settings are working -> some bigger test calcs on server @ClaraBuettner 
- [x] check if minimal examples are working -> @KathiEsterl 
- [x] check if documentation is correct -> @KathiEsterl 
- [x] add release date to whats_new
- [x] replace version number
- [x] don't forget to replace the year and version numbers also on doc/conf.py
- [x] pot. update Wiki on how to release

KathiEsterl:
- [x] review #864 -> @ClaraBuettner 
- [x] test eGo with #863 -> @KathiEsterl 
- [x] depending on the eGo test: review and merge #863 -> @ClaraBuettner

KathiEsterl:
I tested eGo with #856 and committed some minor changes needed for this update. Anyhow, I would recommend to include PR #863 in the new release. Therefore, it is ready for another check by you, @ClaraBuettner . I will already approve it, you can merge it as soon as you had another look into what I changed afterwards :)

PR #867, opened 2026-08-12
Features/fix tests release 11

ClaraBuettner:
These changes were required to run the tests mentioned in #843

### DIFF
--- a/etrago/disaggregate/spatial.py
+++ b/etrago/disaggregate/spatial.py
@@ -803,7 +803,8 @@ class UniformDisaggregation(Disaggregation):
                     if delta > 1e-4:
                         print(
                             "Warning: Sum of disaggregated time series does"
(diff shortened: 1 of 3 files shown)


KathiEsterl:
I checked the code, looks fine to me. Do you want me to run a local test or do you think it is fine like this?

ClaraBuettner:
I think the most critical change is the limit of the pandas version. 
Could you maybe check in your venv which one you are using? 

The code changes were really small, so I guess they don't need to be tested again.

KathiEsterl:
Actually, in my venv I have pandas==2.1.4.

KathiEsterl:
The restriction to pandas<2.0.0 seems to be necessary to conduct the market optimization with pyomo.

KathiEsterl (approved):

PR #869, opened 2026-08-13
Fixes/#868 fix bugs batteries pnommin

### LINKED ISSUE #868: Issues with clustering and disaggregation of ext. storage_units with p_nom_min
The bottom-up approach revealed two problems with the home storage units modelled as a p_nom_min of the battery units:

- p_nom_min of extendable storage units is not considered in the weighting during the spatial clustering
- p_nom_min is not considered in the spatial disaggregation

ClaraBuettner:
Fixes #868

### DIFF
--- a/etrago/analyze/calc_results.py
+++ b/etrago/analyze/calc_results.py
@@ -243,7 +243,7 @@ def calc_investment_cost(self):
     ext_store = network.stores[network.stores.e_nom_extendable]
 
     if not ext_storage.empty:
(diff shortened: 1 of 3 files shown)


ClaraBuettner:
A first local test looks good. 
I will restart the disaggregation for a bigger run. 
But I think, in general, this is ready to be reviewed :)

KathiEsterl:
I checked the code and also ran it locally.
I’d suggest waiting for the results of the larger run. If those look good, I’d say we’re ready to merge.

KathiEsterl (approved):

